Issue #8, opened 2018-10-16
Drop `ishermitian` and `opnorm` requirement for custom operator types.


MSeeker1340:
Currently to use a matrix-free operator we need to implement the `LinearAlgebra.ishermitian` and `LinearAlgebra.opnorm` (or provide a custom norm function). These slows down rapid prototyping and for some sparse types (e.g. `SymTridiagonal`) the default methods are not very efficient.

Since only scalar values are really needed, we should change the methods to not accept the functions themselves but their values with defaults (as suggested in #1).

ChrisRackauckas:
Instead of `opnorm`, could we use the norm of the vector from one application? That was mentioned in https://github.com/Jutho/KrylovKit.jl/issues/22

PR #9, opened 2018-10-17
ExponentialUtilities v1.3

MSeeker1340:
Two main jobs to do:

- [x] Address #8.
- [ ] ~~Write new version of `phiv!` analogous to `expv!` with Saad.~~

Note that I've refactored `expv` into two functions, depending on the choice of `mode`. This is because the code for the two stopping modes are vastly different and they use different keyword arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,13 +50,22 @@ In both cases, `ts` is an array of time snapshots for u, with `U[:,j] ≈ u(ts[j
 
 Apart from keyword arguments that affect the computation of Krylov subspaces (see the Arnoldi iteration section), you can also adjust the timestepping behavior using the arguments. By setting `adaptive=true`, the time step and Krylov subsapce size adaptation scheme of Niesen & Wright is used and the relative tolerance of which can be set using the keyword parameter `tol`. The `delta` and `gamma` parameter of the adaptation scheme can also be adjusted. The `tau` parameter adjusts the size of the timestep (and for `adaptive=true`, the initial timestep). By default, it is calculated using a heuristic formula by Niesen & Wright.
 
+### Support for matrix-free operators
+
+You can use any object as the "matrix" `A` as long as it implements the following linear operator interface:
+
+* `LinearAlgebra.mul!(y, A, x)` (for computing `y = A * x` in place).
+* `Base.size(A, dim)`
+* `LinearAlgebra.opnorm(A, p=Inf)`. If this is not implemented or the default implementation can be slow, you can manually pass in the operator norm (a rough estimate is fine) using the keyword argument `opnorm`.
+* `LinearAlgebra.ishermitian(A)`. If this is not implemented or the default implementation can be slow, you can manually pass in the value using the keyword argument `ishermitian`.
+
 ## Matrix-phi function `phi`
 
 ```julia
 phi(z,k[;cache]) -> [ϕ_0(z),ϕ_1(z),...,ϕ_k(z)]
 ```
 
-Compute ϕ function directly. `z` can both be a scalar or a Matrix. This is used by the caching versions of the ExpRK integrators to precompute the operators.
+Compute ϕ function directly. `z` can both be a scalar or a `AbstractMatrix` (note that unlike the previous functions, you *need* to use a concrete matrix). This is used by the caching versions of the ExpRK integrators to precompute the operators.
 
 Instead of using the recurrence relation, which is numerically unstable, a formula given by Sidje is used [2].
 
--- a/src/krylov_phiv.jl
+++ b/src/krylov_phiv.jl
@@ -32,27 +32,30 @@ employed by setting the kwarg `mode=:error_estimate`.
 
 Compute the expv product using a pre-constructed Krylov subspace.
 """
-function expv(t, A, b; m=min(30, size(A, 1)), tol=1e-7, rtol=√(tol),
-              opnorm=LinearAlgebra.opnorm, cache=nothing, iop=0,
-              mode=:happy_breakdown)
-    Ks = if mode == :happy_breakdown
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -50,13 +50,22 @@ In both cases, `ts` is an array of time snapshots for u, with `U[:,j] ≈ u(ts[j
...
 
+### Support for matrix-free operators
+
+You can use any object as the "matrix" `A` as long as it implements the following linear operator interface:
+
+* `LinearAlgebra.mul!(y, A, x)` (for computing `y = A * x` in place).
+* `Base.size(A, dim)`
+* `LinearAlgebra.opnorm(A, p=Inf)`. If this is not implemented or the default implementation can be slow, you can manually pass in the operator norm (a rough estimate is fine) using the keyword argument `opnorm`.
+* `LinearAlgebra.ishermitian(A)`. If this is not implemented or the default implementation can be slow, you can manually pass in the value using the keyword argument `ishermitian`.
```
ChrisRackauckas: you might want to mention the DiffEqOperator helper @YingboMa made for this. We should add `ishermitian` and `opnorm` to it though.

src/krylov_phiv.jl
```diff
@@ -32,32 +32,36 @@ employed by setting the kwarg `mode=:error_estimate`.
...
+    Ks = arnoldi(A, b; kwargs_arnoldi...)
     w = similar(b)
-    if mode == :happy_breakdown
-        expv!(w, t, Ks; cache=cache)
-    elseif mode == :error_estimate
-        expv!(w, t, A, b,
-              Ks, get_subspace_cache(Ks);
-              atol=tol, rtol=rtol)
-    end
+    expv!(w, t, Ks; cache=cache)
```
ChrisRackauckas: what's this calling?

src/krylov_phiv.jl
```diff
@@ -32,32 +32,36 @@ employed by setting the kwarg `mode=:error_estimate`.
...
+end
+function _expv_ee(t, A, b; m=min(30, size(A, 1)), tol=1e-7, rtol=√(tol),
+    ishermitian::Bool=LinearAlgebra.ishermitian(A))
+    # Error-estimate mode: construction of Krylov subspace and expv! at the same time
+    n = size(A,1)
+    T = promote_type(typeof(t), eltype(A), eltype(b))
+    U = ishermitian ? real(T) : T
+    Ks = KrylovSubspace{T,U}(n, m)
+    w = similar(b)
+    expv!(w, t, A, b, Ks, get_subspace_cache(Ks); atol=tol, rtol=rtol)
```
ChrisRackauckas: this is the `expv!` I'd expect?

ChrisRackauckas:
It's confusing that both of them are called `expv!` with different arguments. It'll be fine if it gets appropriate documentation though.

ChrisRackauckas:
I like the mode choice in the high level allocating version.

jagot:
I think it looks nice!

MSeeker1340:
Did some extensive structural changes. `krylov_expv.jl` is renamed to `krylov_phiv_error_estimate.jl` and the `expv!` function using Saad is renamed to `expv_ee!`. Furthermore, things related to stegr is separated into two parts: those independent of exponentiation (basically `StegrWork`) are moved to a new inner module `Stegr` while those used by `expv!_ee` are merged into `krylov_phiv_error_estimate.jl`.

@jagot would you like to develop `Stegr` into a separate package? Maybe also extend `LinearAlgebra.eigen!` for `SymTridiagonal`s so that user-level code doesn't need to use `stegr!`?

BTW, I'd like to drop the initial plan of implementing `phiv_ee!` since I'm still debating about the details, and this PR already has enough changes. I'll also need to update OrdinaryDiffEq regarding the `opnorm` change after tagging the v1.3 release. @ChrisRackauckas for the update, should I keep backward compatibility (i.e. still allow passing in the `opnorm` function instead of a value) for now so that update on the OrdinaryDiffEq part is easier?

jagot:
In fact, I do _not_ like the change to `expv_ee!`, quite the contrary. As I lined out in #10, I would prefer it to not even have the cache as a separate argument. I think the operator should be tagged with the cache such that the correct algorithm is chosen via dispatch, so you only ever need to write `expv!(b, t, A, x)`.

Regarding `stegr!`, I don't think it should go into a separate package, as it breaks the functionality as-is at the present. Merely implementing `eigen!` will not work since more cache is necessary beyond a tridiagonal matrix. I also don't think we need shun Lapack functions, just look at e.g. BandedMatrices.jl

MSeeker1340:
> In fact, I do not like the change to expv_ee!, quite the contrary. As I lined out in #10, I would prefer it to not even have the cache as a separate argument. I think the operator should be tagged with the cache such that the correct algorithm is chosen via dispatch, so you only ever need to write expv!(b, t, A, x)

Fair point.

> Regarding stegr!, I don't think it should go into a separate package, as it breaks the functionality as-is at the present.

Since we're just extending `stegr!` with `StegrWork`, how is this breaking the current funtionality? The original `stegr!` isn't changed after all.

> Merely implementing `eigen!` will not work since more cache is necessary beyond a tridiagonal matrix.

Sorry I'm not making things clear. What I suggest is to add

```julia
import LinearAlgebra: eigen!

function eigen!(A::SymTridiagonal{T}, sw::StegrWork) where {T <: Real}
  stegr!(A.dv, A.de, sw)
  ... # wrap things in an Eigen object
end
```

The intention is that, if you're a power user and want a faster `eigen!`, you can just allocate a `StegrWork` and then proceed to use `eigen!` pretty much the same way as before, without having to switch to `stegr!` in the user code.

PR #15, opened 2019-02-28
make arnoldi work for complex hermitian

Roger-luo:
Hey, guys. I found this was not working for complex hermitian, just a small fix to make it work for hermitians.

### DIFF
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -82,7 +82,7 @@ Numerical Mathematics and Advanced Applications-ENUMATH 2013 (pp. 345-353).
 Springer, Cham.
 """
 function arnoldi(A, b; m=min(30, size(A, 1)), kwargs...)
-    Ks = KrylovSubspace{eltype(b)}(length(b), m)
+    Ks = KrylovSubspace{eltype(b), real(eltype(b))}(length(b), m)
     arnoldi!(Ks, A, b; m=m, kwargs...)
 end
 
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -82,7 +82,8 @@ Numerical Mathematics and Advanced Applications-ENUMATH 2013 (pp. 345-353).
 Springer, Cham.
 """
 function arnoldi(A, b; m=min(30, size(A, 1)), kwargs...)
-    T = eltype(b)
+    TA, Tb = eltype(A), eltype(b)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
The code itself looks fine, I just don't know the subtle details here of whether this algorithm should be allowed? @jagot would be the guy for that.

Roger-luo:
@jagot have reviewed it in slack. I guess it's fine then. I'll try to solve #16 later.

jagot:
I think it should br OK, but let me think about it one more round.

Roger-luo:
I think this should be tweaked a little bit. Since currently `T` follows `eltype(b)`, but `A` can be complex and `b` can be real, then the eltype `T` for `KrylovSubSpace` should be complex as well.

currently, this only works when `b` is complex.

Although, it's might be rare to use a complex `A` and a real `b` in `expv` but I guess this is allowed.

Roger-luo:
I try to use this function (no in-place function like `lmul!`) for complex `t`, but anyone know why this is not correct when `A` is a hermitian. It works when `t` is real, but not working when `t` is complex.

```julia
function expv!(w::AbstractVector{Complex{Tw}}, t::Complex{Tt}, Ks::KrylovSubspace{B, T, U};
                cache=nothing) where {Tw, Tt, B, T, U}
    m, beta, V, H = Ks.m, Ks.beta, getV(Ks), getH(Ks)
    @assert length(w) == size(V, 1) "Dimension mismatch"
    if cache == nothing
        cache = Matrix{U}(undef, m, m)
    elseif isa(cache, ExpvCache)
        cache = get_cache(cache, m)
    else
        throw(ArgumentError("Cache must be an ExpvCache"))
    end
    copyto!(cache, @view(H[1:m, :]))
    if ishermitian(cache)
        # Optimize the case for symtridiagonal H
        expHe = F.vectors * exp.(t * F.values) .* @view(F.vectors[1, :])
    else
        expH = exp!(t * cache)
        expHe = @view(expH[:, 1])
    end
    lmul!(beta, mul!(w, @view(V[:, 1:m]), expHe)) # exp(A) ≈ norm(b) * V * exp(H)e
end
```

Roger-luo:
OK, I fixed #16 now. All kinds of complex variables are working now. I guess you have to review this again: @ChrisRackauckas @jagot

PR #20, opened 2019-03-14
WIP: Non-allocating `exp!`

MSeeker1340:
I have translated the `exp!` function in `LinearAlgebra` to a non-allocating version (temporarily named `_exp!`). The next step is to replace the `exp!` occurrences in the package to `_exp!` and then remove the former.

In cases where the matrices for which `exp!` is called are small (i.e. small number of Krylov iterations) I suspect there won't be too much improvement (though adding them up can still probably save some noticeable time). The improvement should be more visible for larger Krylov subspaces.

### DIFF
--- a/src/ExponentialUtilities.jl
+++ b/src/ExponentialUtilities.jl
@@ -1,7 +1,7 @@
 module ExponentialUtilities
 using LinearAlgebra, SparseArrays, Printf
-using LinearAlgebra: exp!, BlasInt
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
This seems like a nice thing to have. Any chance it will get finished? It looks like it just needs to be setup to be used now?

ChrisRackauckas:
I wonder if this library needs `@..` to be competitive?

PR #26, opened 2019-07-06
Remove scaling tolerance by opnorm

YingboMa:
As the stiffness of Jacobian increases, the opnorm increases. Thus, scaling tolerance by opnorm would be detrimental to the accuracy of stiff equations.


YingboMa:
Before:
Convergence plot of POLLU:
convergence_before
Benchmark plot of POLLU:
benchmark_before


After:
Convergence:
convergence_after
Benchmark:
benchmark_after

PR #28, opened 2019-12-09
Add generic exp function.

MasonProtter:
This adds a generic, pure julia `exp_generic` function that should work on a wide variety of data types.

### DIFF
--- a/src/ExponentialUtilities.jl
+++ b/src/ExponentialUtilities.jl
@@ -22,6 +22,6 @@ include("StegrWork.jl")
 include("krylov_phiv_error_estimate.jl")
 
 export phi, phi!, KrylovSubspace, arnoldi, arnoldi!, lanczos!, ExpvCache, PhivCache,
-    expv, expv!, phiv, phiv!, kiops, expv_timestep, expv_timestep!, phiv_timestep, phiv_timestep!,
+    expv, expv!, exp_generic, phiv, phiv!, kiops, expv_timestep, expv_timestep!, phiv_timestep, phiv_timestep!,
     StegrCache, get_subspace_cache
 end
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Should this and `exp!` merge in some way?

MasonProtter:
There’s honestly not a lot of overlapping functionality between them. One thing I would like to do is write a `evalpoly!` that uses the 5 argument `mul!` instead of `muladd`. After that, I think `exp_generic` should be around as fast.

ChrisRackauckas:
I see, they are two different algorithms, so both useful in their own sense. It would be good to get docs on it.

jagot:
I just came across this, and it's a nice addition! However, instead of `BigFloat`s and `factorial`s, could not the rational polynomial be evaluated using the gamma function? For instance, in the error estimate version of the Krylov iterations, the subspace exponential is computed for every Krylov iteration (of which there could be many for each time step):
https://github.com/JuliaDiffEq/ExponentialUtilities.jl/blob/4ec34f33f0a7584e73e5f7d378506bf58d2dfa05/src/krylov_phiv_error_estimate.jl#L77-L78

The next thing, but I think it would be covered by a 5-arg version, is to only evaluate the action of the subspace exponential, not instantiating the matrix.

MasonProtter:
> However, instead of BigFloats and factorials, could not the rational polynomial be evaluated using the gamma function? For instance, in the error estimate version of the Krylov iterations, the subspace exponential is computed for every Krylov iteration (of which there could be many for each time step)

I’m not sure I understand. I just used `factorial` and *`BigInt`* because the factorials seems computed at compile time, so speed doesn’t really matter and I wanted people to have the option of going arbitrarily high in Padé series.

jagot:
That's very neat! I must confess I missed that point...

PR #36, opened 2020-06-11
generalize function definitions to support cuda

### LINKED ISSUE #27: support CUDA arrays?
I tried to port CuArrays `CUSPARSE` module to ExponentialUtilities  to utilize GPU devices. I removed the array type restriction from `KrylovSubspace`, but still not working.

Wondering how much effort it is to write a GPU compatible `expmv` and does it worth a try?

Roger-luo:
I generalized some definitions to support CUDA, which closes #27 . but to make `expv!` work on CUDA, one needs this patch: 

https://github.com/JuliaGPU/CUDA.jl/pull/217

and this function

```jl
function ExponentialUtilities.expv!(w::CuVector{Tw}, t::Real, Ks::KrylovSubspace{T, U};
               cache=nothing, dexpHe::CuVector = CuVector{U}(undef, Ks.m)) where {Tw, T, U}
    m, beta, V, H = Ks.m, Ks.beta, getV(Ks), getH(Ks)
    @assert length(w) == size(V, 1) "Dimension mismatch"
    if cache == nothing
        cache = Matrix{U}(undef, m, m)
    elseif isa(cache, ExpvCache)
        cache = get_cache(cache, m)
    else
        throw(ArgumentError("Cache must be an ExpvCache"))
    end
    copyto!(cache, @view(H[1:m, :]))
    if ishermitian(cache)
        # Optimize the case for symtridiagonal H
        expHe = F.vectors * (exp.(lmul!(t,F.values)) .* @view(F.vectors[1, :]))
    else
        lmul!(t, cache); expH = cache
        _exp!(expH)
        expHe = @view(expH[:, 1])
    end

    copyto!(dexpHe, expHe)
    lmul!(beta, mul!(w, @view(V[:, 1:m]), dexpHe)) # exp(A) ≈ norm(b) * V * exp(H)e
end
```

since this package is quite light-weight, I suppose it'd be better to have a separate package to support CUDA (so you don't have to download the entire CUDA when installing this for just CPU).

to give whoever interested in this a feeling about the acceleration, on our AWS machine with Tesla M60 GPU, I tested with A of size 10_000 x 10_000, b of size 10_000, t= 1e-2, for different method and arrays, this gives the following acceleration

Dense: 4.46x faster
Dense Hermitian: 2.75x faster
sparse(A), dense(b): 14.17x faster
sparse(A), dense(b) (Hermitian): 3.27x faster

Tho, it seems strange to me that Hermitian has lower acceleration rate (and is slower than the non-hermitian method), but this should give a speedup in general for many applications. I'll investigate this later maybe.

For folks interested in the implementation of CUDA, could refer to this gist for now: https://gist.github.com/Roger-luo/e7704108157ddea579e140354c4ba6df

### DIFF
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -40,10 +40,9 @@ end
 
 KrylovSubspace{T}(args...) where {T} = KrylovSubspace{T,T}(args...)
 
-
 getV(Ks::KrylovSubspace) = @view(Ks.V[:, 1:Ks.m + 1])
 getH(Ks::KrylovSubspace) = @view(Ks.H[1:Ks.m + 1, 1:Ks.m+!iszero(Ks.augmented)])
-function Base.resize!(Ks::KrylovSubspace{B,T,U}, maxiter::Integer) where {B,T,U}
+function Base.resize!(Ks::KrylovSubspace{T,U,B}, maxiter::Integer) where {B,T,U}
     isaugmented = !iszero(Ks.augmented)
     V = Matrix{T}(undef, size(Ks.V, 1), maxiter + 1)
     H = fill(zero(U), maxiter + 1, maxiter + isaugmented)
@@ -204,7 +203,7 @@ end
 
 Non-allocating version of `arnoldi`.
 """
-function arnoldi!(Ks::KrylovSubspace{B, T1, U}, A::AT, b;
+function arnoldi!(Ks::KrylovSubspace{T1, U, B}, A::AT, b;
                   tol::Real=1e-7, m::Int=min(Ks.maxiter, size(A, 1)),
                   ishermitian::Bool=LinearAlgebra.ishermitian(A isa Tuple ? first(A) : A),
                   opnorm=nothing, iop::Int=0,
@@ -277,7 +276,7 @@ realview(::Type{R}, V::AbstractVector{R}) where {R} = V
 A variation of `arnoldi!` that uses the Lanczos algorithm for
 Hermitian matrices.
 """
-function lanczos!(Ks::KrylovSubspace{B, T1, U}, A::AT, b;
+function lanczos!(Ks::KrylovSubspace{T1, U, B}, A::AT, b;
                   tol=1e-7, m=min(Ks.maxiter, size(A, 1)),
                   opnorm=nothing,
                   init::Int=0, t::Number=NaN, mu::Number=NaN, l::Int=-1) where {B, T1 <: Number, U <: Number, AT}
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
What makes that `expv!` function non-generic and require dispatching directly for CuArrays? I don't see what's explicitly calling CUDA in that function

ChrisRackauckas:
If we need to we can `@requires`, but I don't see what's so specific about that dispatch that the type signatures can't just be more generic and replace what's currently in the library.

Roger-luo:
@ChrisRackauckas it's not about dispatching, it's about the actual implementation. One should not allocate `Ks.H` on GPU, since `Ks.H` is of size `m` which is by default 30, thus not very large. However, in `expv!` one needs to calculate the (in-place version) matrix `exp` and could involve calling to `eigen` which is not supported (and can be slower since this is small matrix eigen decomposition)

the current implementation will copy the CPU `Ks.H` to GPU in the end.

ChrisRackauckas:
And why can't the CPU version be made to be the same as this one? An efficiency reason? (I'm asking this without looking at the current one in detail)

Roger-luo:
@ChrisRackauckas there is only one device for the CPU version, you don't need to copy between devices.

Roger-luo:
This line is the only difference in GPU version (but this causes the interface to accept a GPU buffer)

```jl
copyto!(dexpHe, expHe)
```

could think about providing an extra API to let GPU version to overload (so the GPU version doesn't need to copy the rest of the code)

ChrisRackauckas:
I'll let @YingboMa take it from here. I'm fine with even just using `@requires`, though something a bit more elegant is probably possible

ChrisRackauckas:
I'll merge and do it.

ChrisRackauckas:
Thanks!

ChrisRackauckas:
Actually @Roger-luo could you provide a GPU test? We'll get GPU CI setup here, so just make it a separate file that just does some GPU stuff so we don't accidentally break it.

Roger-luo:
@ChrisRackauckas sure, I think I'm going to do another PR to refactor the API a bit to remove duplication, the current implementation (the GPU one and the complex value one I PR a while ago copied a bunch of code from the real CPU version). I'll add CUDA support after that.

ChrisRackauckas:
I already added in https://github.com/SciML/ExponentialUtilities.jl/pull/37, so you can take it from there. I just wanted to make sure the working solution wasn't lost.

Issue #41, opened 2020-08-15
Avoid generated function in generic_exp


andreasnoack:
The current implementation relies on methods being defined for special number types before https://github.com/SciML/ExponentialUtilities.jl/blob/fce7ae64e8507baf7ce98d268e46478c7a1540ab/src/exp.jl#L127-L135 is being called, see item 4 in https://docs.julialang.org/en/v1/manual/metaprogramming/#Generated-functions-1. If not then you'll hit a world age error. E.g.

```julia
julia> using ExponentialUtilities

julia> f = t -> exp_generic(t)
#1 (generic function with 1 method)

julia> f(0.3)
1.3498588075760032

julia> using ForwardDiff

julia> ForwardDiff.derivative(f, 0.2)
ERROR: MethodError: no method matching ForwardDiff.Dual{ForwardDiff.Tag{var"#1#2",Float64},Float64,1}(::Int64)
The applicable method may be too new: running in world age 27795, while current world is 27806.
```

PR #43, opened 2020-08-16
Work around world age error for Duals in exp_generic

### LINKED ISSUE #42: generic_exp fails in zero
```julia
julia> ExponentialUtilities.exp_generic(0.0)
ERROR: InexactError: trunc(Int64, -Inf)
Stacktrace:
 [1] trunc at ./float.jl:703 [inlined]
 [2] ceil at ./float.jl:365 [inlined]
 [3] exp_generic(::Float64, ::Val{13}) at /Users/andreasnoack/.julia/dev/ExponentialUtilities/src/exp.jl:119
 [4] exp_generic(::Float64) at /Users/andreasnoack/.julia/dev/ExponentialUtilities/src/exp.jl:118
 [5] top-level scope at REPL[64]:1
```

andreasnoack:
and fix `exp_generic(0.0)`.

Closes #42 but not really #41. It's more a temporary workaround than an actual fix. I'm not sure how to fix the problem in full generality.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -37,6 +37,11 @@ using ForwardDiff
     @test ForwardDiff.derivative(exp_generic, 0.1) == exp_generic(0.1)
 end
 
+@testset "Issue 42" begin
+    @test exp_generic(0.0) == 1
+    @test ForwardDiff.derivative(exp_generic, 0.0) == 1
+end
+
 @testset "Phi" begin
     # Scalar phi
     K = 4
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,19 +14,27 @@ using ExponentialUtilities: getH, getV, _exp!
     @test A2 ≈ expA2
 end
 
-@testset "exp_generic" begin 
+@testset "exp_generic" begin
     for n in [5, 10, 30, 50, 100, 500]
         M = rand(n, n)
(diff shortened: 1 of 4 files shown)


andreasnoack:
Please hold off a bit. I need to define the zero branch differently since
```
julia> ForwardDiff.derivative(exp_generic, 0.0)
0.0
```

andreasnoack:
Fixed it so I believe this one should be good to go.

test/runtests.jl
```diff
@@ -14,21 +14,34 @@ using ExponentialUtilities: getH, getV, _exp!
...
-        
+
         N = randn(n, n)
         @test exp(N) ≈ exp_generic(N)
 
         exp(n) ≈ exp_generic(n)
     end
 end
 
+# The issue was only triggered if exp_generic had been called before loading
```
vtjnash, line 32: This should be incorrect. What matters is the load order, while it's a bug if the call order is observable.

ChrisRackauckas: how'd you find this? 😆 I think we can just straight avoid the issue for the most common case (https://github.com/SciML/ExponentialUtilities.jl/pull/45) and handle the more general thing later.

vtjnash: Someone put out the bat signal 🙃

andreasnoack:
Superseded by #45

PR #45, opened 2020-08-16
Hard code 13 term Pade approximate

ChrisRackauckas:
Fixes https://github.com/SciML/ExponentialUtilities.jl/issues/42 and works around https://github.com/SciML/ExponentialUtilities.jl/issues/41 for the standard case.

@andreasnoack can you test this on Pumas?


andreasnoack:
This works for case that used to fail in Pumas.

Issue #40, opened 2020-07-24
Add ChainRules rules


sethaxen:
From Slack:
@sethaxen:
> Does ExponentialUtilities.jl play well with AD packages, in particular Zygote?

@ChrisRackauckas:
> not fully with Zygote
> it'll need adjoints
> since it's doing a lot of scalar stuff
> it's writing the kernels directly
> the adjoints are easy though

I in particular need adjoints for `expv`. Zygote currently has an adjoint rule for `exp(::AbstractMatrix)` and `exp(::Hermitian)` using the eigendecomposition. I imagine though there's a better way to implement the adjoints for `expv` by looking at the underlying algorithm (I have not).

ChrisRackauckas:
There should be ways to do this without defining the Jacobian. `expv` is the solution to the linear ODE, so the adjoint of the ODE can be should be able to be used to derive the expression in terms of the adjoint, which IIRC should just be:

du = Au -> dlambda -> lambda'*A

which means the adjoint should just be expv(A,delta').

sethaxen:
For `λ = expv(t, A, u)` and adjoint of `λ`, `Δλ`, the adjoint of `u` should I think be `∂u = expv(t, A', Δλ)`, which is quite nice.
We also need the adjoints `∂t` and `∂A`, which will take more thought.

sethaxen:
Especially since `A` doesn't even need to be a matrix, right? I don't think we'll be able to support all types for a custom adjoint on `A` then, just  at the most `AbstractMatrix`es.

ChrisRackauckas:
Yeah, the difficult thing will be supporting something that's not concrete, since then it can't adjoint. But then that's just defined as the reverse mode of the function `f(u) = A*u`, so I think it can work out, it'll just be more complicated in code.

Those again would come from this derivation. You might want to read https://diffeq.sciml.ai/stable/extras/sensitivity_math/ or the supplemental of https://arxiv.org/abs/2001.04385 . Specifically, the `∂A` term is given by an integral over the Legrange multiplier term. Coincidentally, the `phiv` values used in the exponential integrators are these integrals, so the adjoint can probably be written as just a calculation of `phi_1`. I think it's like phiv(t, A', Δλ) + reversemode(A) kind of thing (in pseudocode, off the top of my head so maybe missing a detail somewhere). 

[...]

PR #51, opened 2020-12-20
[WIP] Add ChainRules differentiation rules for expv

sethaxen:
Following #40, this PR implements forward- and reverse-mode rules for `expv`, adding compatibility with Zygote and perhaps making ForwardDiff2's differentiation of these rules faster.

This is still marked as a WIP until I figure out sensitivities of the matrix `A`.

### DIFF
--- a/src/krylov_phiv_chainrules.jl
+++ b/src/krylov_phiv_chainrules.jl
@@ -1,0 +1,42 @@
+function ChainRulesCore.frule((_, Δt, ΔA, Δb), ::typeof(expv), t, A, b; kwargs...)
+    w = expv(t, A, b; kwargs...)
+    ∂w = similar(w)
+    mul!(∂w, A, w)
+    ∂w .*= Δt
+    if !isa(Δb, AbstractZero)
+        ∂w .+= expv(t, A, Δb; kwargs...)
+    end
+    # TODO: handle ΔA
+    ΔA isa AbstractZero || error("ΔA currently cannot be pushed forward")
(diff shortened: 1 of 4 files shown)


src/krylov_phiv_chainrules.jl
```diff
@@ -0,0 +1,42 @@
...
+        ∂t = Thunk() do
+            t̄ = A isa AbstractMatrix ? conj(dot(Δw, A, w)) : dot(mul!(similar(w), A, w), Δw)
+            return t isa Real ? real(t̄) : t̄
+        end
+        # TODO: handle ∂A
+        ∂A = @thunk error("Adjoint wrt A not yet implemented")
+        ∂b = Thunk() do
+            # using similar is necessary to ensure type-stability
+            b̄ = similar(b)
+            _copyto!(b̄, expv(t', A', Δw; kwargs...))
```
sethaxen, line 26: `adjoint` is currently not part of the required interface for generic `A`. How should we handle this?
- Make it a required part of the interface
- Make it an optional requirement for `A` (i.,e. "if you want to work with Zygote, you also need to define `adjoint`")
- restrict this `rrule` to `A::AbstractMatrix` (this may be necessary anyways, as for non-matrix `A`, its differential type is unknown.).

ChrisRackauckas: >"if you want to work with Zygote, you also need to define adjoint"

I think that's entirely reasonable.

PR #55, opened 2021-04-09
Faster exp_pade_p(::Matrix, ::Val{13}, ::Val{13})

chriselrod:
Reuses memory to cut down on allocations and keep the memory hotter in the cache (depending on size of the arrays).


chriselrod:
I made the signature `::Matrix` because it's mutating.
We could use something like `ArrayInterface.ismutable`.

ChrisRackauckas:
I think it should use `ArrayInterface.ismutable` there, but this is still an improvement.

PR #60, opened 2021-04-12
Unroll based off of sizeof(eltype(C))

chriselrod:
For huge duals, we probably don't want to unroll much at all.
For small duals, unrolling more helps.

The intention is to try and keep up with `SMatrix` performance, but while this does well in microbenchmarks, in practice it is still several times slower in the real workload, with 80% GC time vs 15% GC time for the SMatrix version. =/


YingboMa (approved):

YingboMa:
Looks like the tests need an update, too.

Issue #61, opened 2021-07-13
Complex valued expv! gives MethodError in some cases


Lazersmoke:
```julia
A = Diagonal(1 .* ones(ComplexF64,8))
b = ones(ComplexF64,8)

Ks = KrylovSubspace{ComplexF64,ComplexF64}(8,8)
arnoldi!(Ks,A,b) # After this like, Ks.H = [ 1 - epsilon + 0i ; epsilon + 0i ]
out = Vector{ComplexF64}(undef,8)
expv!(out,1.0,Ks) # Causes error
```
Since `H[1:1,1]` is hermitian, this branch goes through:
https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/krylov_phiv.jl#L82-L85
but the efficient `eigen!` for SymTridiagonal matrices only works for reals (see https://github.com/JuliaLang/julia/commit/fc4c69de3935206589d25c251f5423cee1601726) so you get this error:

```
ERROR: LoadError: MethodError: no method matching eigen!(::SymTridiagonal{ComplexF64, Vector{ComplexF64}})
Closest candidates are:
  eigen!(!Matched::SymTridiagonal{var"#s832", V} where {var"#s832"<:Union{Float32, Float64}, V<:AbstractVector{var"#s832"}}) at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\LinearAlgebra\src\tridiag.jl:280
  eigen!(!Matched::SymTridiagonal{var"#s832", V} where {var"#s832"<:Union{Float32, Float64}, V<:AbstractVector{var"#s832"}}, !Matched::UnitRange) at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\LinearAlgebra\src\tridiag.jl:283
  eigen!(!Matched::SymTridiagonal{var"#s832", V} where {var"#s832"<:Union{Float32, Float64}, V<:AbstractVector{var"#s832"}}, !Matched::Real, !Matched::Real) at C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\LinearAlgebra\src\tridiag.jl:288
  ...
```

In the `expv!` code for complex `t`, there is this line:
https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/krylov_phiv.jl#L116
[...]

PR #62, opened 2021-07-13
Add cast to real for Gram-Schmidt coefficients

Lazersmoke:
According to this line, in the hermitian case, the actual coefficients are supposed to be real:
https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/arnoldi.jl#L32
but the matrix type can be complex. `eigen!` is not implemented for complex-valued
symmetric tridiagonal matrices in LinearAlgebra, so this throws a method
error if the cache is complex hermitian.

This needs verification to ensure this is actually correct, but making this change makes the code work for my application at least.

Closes https://github.com/SciML/ExponentialUtilities.jl/issues/61

### DIFF
--- a/src/krylov_phiv.jl
+++ b/src/krylov_phiv.jl
@@ -82,7 +82,7 @@ function expv!(w::AbstractVector{Tw}, t::Real, Ks::KrylovSubspace{T, U};
     copyto!(cache, @view(H[1:m, :]))
     if ishermitian(cache)
         # Optimize the case for symtridiagonal H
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
That seems sketch and possibly not true in general. @YingboMa ?

Lazersmoke:
Turns out you should not do this. The *diagonal* Gram-Schmidt coefficients (so, `H[i,i]` for each `i`) are real, but not the off-diagonal ones.

PR #64, opened 2021-08-17
New implementation of `_exp!` based on generated code

### LINKED ISSUE #63: One multiplication too many in `_exp!`
Analogous to the PR: https://github.com/JuliaLang/julia/pull/40668 I believe the `_exp!`-function can be implemented with one multiplication less in several cases. More precisely, these lines can be implemented with one multiplication less:

https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/exp.jl#L41-L49 

Since when `k=1`, `P` is identity.

jarlebring:
This is a new implementation of non-allocating matrix exponential `_exp!`. This implementation is better in terms of efficiency since it fixes #63, but also makes more use of dot fusion.  

```julia-repl
julia> using BenchmarkTools, LinearAlgebra, Random
julia> Random.seed!(0);
julia> n=1024*2;
julia> A=randn(n,n);
julia> A=0.85*A/opnorm(A,1);
julia> cache=(similar(A),similar(A),similar(A),similar(A),similar(A));
julia> Eold=@btime ExponentialUtilities._baseexp!(copy($A),caches=$cache);
  1.614 s (7 allocations: 32.03 MiB)
julia> Enew=@btime ExponentialUtilities._exp!(copy($A),caches=$cache);
  1.415 s (53 allocations: 32.04 MiB)
julia> norm(Eold-Enew)
1.696649934880783e-15
```

The implementation is based on generated code.  It is generated using the package https://github.com/matrixfunctions/GraphMatFun.jl and the files `exp_XX.jl` in `exp_generated` are generated by the script https://github.com/jarlebring/ExponentialUtilities.jl/blob/master/src/exp_generated/generate_exp_code.jl


src/exp2.jl
```diff
@@ -0,0 +1,66 @@
...
+    n = LinearAlgebra.checksquare(A)
+    nA = opnorm(A,1);
+
+    # Maybe to balancing
+    if do_balancing
+        ilo, ihi, scale = LAPACK.gebal!('B', A)    # modifies A
+    end
+
+    # Select how many multiplications to use
+    rhov=    [0.015; 0.25; 0.95; 2.1; 5.4];
```
YingboMa, line 20: This could be moved outside and become a global constant, so that Julia won't allocate it each time when this function is invoked.

test/runtests.jl
```diff
@@ -15,6 +15,20 @@ using ForwardDiff, StaticArrays
     @test A2 ≈ expA2
 end
 
+@testset "Exp2" begin
+    n = 100
+    A = randn(n, n)
+    expA = exp(A)
+    _exp2!(A)
```
YingboMa, line 22: Remember to import it.

YingboMa (approved):
Haha, this is probably good for precompiles. @ChrisRackauckas will like this :-p

src/exp_generated/exp_10.jl
```diff
@@ -0,0 +1,108 @@
+using LinearAlgebra
+
+@inline function exp_10(A)
+    T=promote_type(eltype(A),Float64)
+    A_copy=similar(A,T); A_copy .= A;
```
ChrisRackauckas, line 5: copyto!

src/exp2.jl
```diff
@@ -0,0 +1,66 @@
...
+    else
+        thiscache = caches;
+    end
+
+
+    # Evaluate
+    exp_fun=Symbol("exp_$(i)!")
+    s=:($(exp_fun)($thiscache,$A));
+    X=eval(s)::typeof(A) # Evaluate and maintain type stability
+
```
ChrisRackauckas, line 43: no need for eval here, just branch.

jarlebring: I don't understand. What do you mean by branch?

src/exp2.jl
```diff
@@ -0,0 +1,66 @@
+# Import the generated code
+for i=1:15
+    include("exp_generated/exp_$i.jl")
+end
+
+function getmem(cache,k)
+    return cache[k-1];
+end
+
+function _exp2!(A; caches=nothing, do_balancing = A isa StridedMatrix)
```
YingboMa, line 10: I assume this is an internal function that shouldn't be called besides testing?

jarlebring: I view `_exp2!` equivalent to `_exp!` or a replacement thereof. When we are done comparing `_exp!` and `_exp2!`, we could consider renaming `_exp!` <- `_exp2!`. The function `_exp!` is currently not exported.

src/exp2.jl
ChrisRackauckas: `if` statements. Relying on `eval` is just a very indirect if statement that will have more overhead

jarlebring: You mean like this? 
```
if (nA < rv[1])
   exp_1!(cache,A);
elseif (nA < rv[2]) 
   exp_2!(cache,A);
...
else 
   exp_15!(cache,A);
end
```
It's more code, but if you're saying the `eval` call is a problem it's okay with me.

ChrisRackauckas: yeah, like that. Eval in Julia is dynamic and has to invoke the compiler, so there's a lot of issues there (it cannot optimize across the barrier, has a good amount of overhead, won't precompile, etc.) so it's best avoided in numerical codes. A big set of if statements is more code but less compute instructions.

src/exp2.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    # Select how many multiplications to use
+    rhov=    [0.015; 0.25; 0.95; 2.1; 5.4];
+    # Number of memslots needed (beside A)
+    memslots=[3   ;    4;    5;   6;   5];
+    for s=1:8
+        push!(rhov, rhov[end]*2);
+        push!(memslots,memslots[end]);
+    end
+    i = findfirst(nA .< rhov)
```
YingboMa, line 27: This allocates. Try
```suggestion
    i = findfirst(nA < rho for rho in rhov)
```

src/exp2.jl
YingboMa:
```julia
help?> Base.@nif
  @nif N conditionexpr expr
  @nif N conditionexpr expr elseexpr

  Generates a sequence of if ... elseif ... else ... end statements. For example:

  @nif 3 d->(i_d >= size(A,d)) d->(error("Dimension ", d, " too big")) d->println("All OK")

  would generate:

  if i_1 > size(A, 1)
      error("Dimension ", 1, " too big")
  elseif i_2 > size(A, 2)
      error("Dimension ", 2, " too big")
  else
      println("All OK")
  end
```
Can help you if you want elegant code.

jarlebring: hmmm.. thanks. interesting. Elegant yes. Might be more difficult to read... at least for me.  :confused:

src/exp2.jl
```diff
@@ -0,0 +1,90 @@
...
+        # No cache. Allocate Matrix objects
+        TT=eltype(A);
+        thiscache=[Matrix{TT}(undef,n,n) for i=1:memslots[i]];
+    else
+        thiscache = caches;
+    end
+
+
+    # Evaluate
+    if (nA < rhov[1])
```
chriselrod, line 40: Could make it a `Val` arg, and then something like
```julia
Base.Cartesian.@nif 13 d -> nA < rhov[d] d -> X = __exp!(thiscache, A, Val(d)) d -> X = __exp!(thiscache, A, Val(d))
```
This produces
```julia
:(if nA < rhov[1]
      X = __exp!(thiscache, A, Val(1))
  else
      if nA < rhov[2]
          X = __exp!(thiscache, A, Val(2))
      else
          if nA < rhov[3]
              X = __exp!(thiscache, A, Val(3))
          else
              if nA < rhov[4]
                  X = __exp!(thiscache, A, Val(4))
              else
                  if nA < rhov[5]
                      X = __exp!(thiscache, A, Val(5))
                  else
                      if nA < rhov[6]
                          X = __exp!(thiscache, A, Val(6))
                      else
                          if nA < rhov[7]
                              X = __exp!(thiscache, A, Val(7))
                          else
                              if nA < rhov[8]
                                  X = __exp!(thiscache, A, Val(8))
                              else
                                  if nA < rhov[9]
                                      X = __exp!(thiscache, A, Val(9))
                                  else
                                      if nA < rhov[10]
                                          X = __exp!(thiscache, A, Val(10))
                                      else
                                          if nA < rhov[11]
                                              X = __exp!(thiscache, A, Val(11))
                                          else
                                              if nA < rhov[12]
                                                  X = __exp!(thiscache, A, Val(12))
                                              else
                                                  X = __exp!(thiscache, A, Val(13))
                                              end
                                          end
                                      end
                                  end
                              end
                          end
                      end
                  end
              end
          end
      end
  end)
```

jarlebring: And then add a function `__exp!`? Looks quite complicated. I'm slightly in favor of the current hard-coded if-statements.

src/exp_generated/exp_10.jl
jarlebring: Should be fixed now.

jarlebring:
Any specific thoughts about using generated code? 

The functions `exp_6!,`... `exp_13!` are very similar, and one could reduce them to one by using the identity `exp_7!(A)=exp_6!(A/2)^2` etc. It would probably require more computation (a scaling of the matrix) but the added computation might be acceptable to get less code.

ChrisRackauckas:
It's fine, used the generated code.

src/exp2.jl
chriselrod: It'd be replacing `exp12(x,y) = ...` with `__exp(x, y, ::Val{12}) = ...`.

ChrisRackauckas:
Test failure looks real

jarlebring:
I believe `E=_exp!(A)` writes the solution into `A`, whereas `E=_exp2!(A)` has destroyed the contents of `A` but it does not contain the solution. When I run tests locally (julia 1.7-dev), this seems to be the issue. The  CI seems to complain about something related to `findfirst`...

src/exp2.jl
jarlebring: Okay. I understand. Rather than many different functions with different names, one could selected it with dispatch. It's doable with the code generation. Advantages?

chriselrod: Mostly that it's fewer lines of code, and easier to cover with tests.
I agree that it is questionable whether it's actually easier to read.
It can be broken up to make it a little more clear:
```julia
X = Base.Cartesian.@nif 13 d -> begin
    nA < rhov[d]
end d -> begin # if condition
    __exp!(thiscache, A, Val(d))
end d -> begin # fallback (d == 13)
    __exp!(thiscache, A, Val(d))
end
```

jarlebring: That looks okay to read. Done.

jarlebring:
Thoughts about replacing `_exp` with `_exp2`? Any important users/use-cases that should be checked?

ChrisRackauckas:
Check whether the generated kernels for CUDA work (probably not without changing the LAPACK calls?).

The main use cases are the exponential integrators https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/krylov_phiv.jl . 

It might be interesting to run the exprk benchmark in https://benchmarks.sciml.ai/html/MOLPDE/allen_cahn_fdm_wpd.html and see how that's effected.

jarlebring:
Currently, CI is complaining about `mul!`. This a version issue. The code does identity addition with https://github.com/JuliaLang/julia/pull/40731, which is not in the CI version of julia (1.6.2). If we want support for 1.6.2, I could change it by basically including the diagonal add in 40731 as a function `uniformscaling_add!`.

jarlebring:
> Check whether the generated kernels for CUDA work (probably not without changing the LAPACK calls?).

If we disable balancing, there is only one LAPACK call the linear solve `LAPACK.gesv!`. What should the be in a CUDA setting?

src/exp2.jl
jarlebring: I had to revert this since it caused CI to fail... julia version issue?

src/exp2.jl
```diff
@@ -3,9 +3,17 @@ for i=1:13
     include("exp_generated/exp_$i.jl")
 end
 
-function getmem(cache,k)
+function getmem(cache,k) # Called from generated code
     return cache[k-1];
 end
+function ldiv_for_generated!(C,A,B) # C=A\B. Called from generated code
```
ChrisRackauckas, line 9: why not just use `ldiv!`?

jarlebring: I think `ldiv!` is not intended for matrices but only factorizations. From `?ldiv!`:
```
  ldiv!(A, B)

  Compute A \ B in-place and overwriting B to store the result.

  The argument A should not be a matrix. Rather, instead of matrices it should be a
```

ChrisRackauckas: just do the `lu!` first?

jarlebring: Not sure exactly how `lu!`+`ldiv!` is different from `LAPACK.gesv!`. The `gesv!` does an LU and ldiv in fortran code, but maybe it does something more. They do have very small differences in CPU-timing and allocations. Both `_exp!` and `Base.exp` use the LAPACK-call.  

Would `lu!`+`ldiv!` make things easier in the CUDA-setting?

ChrisRackauckas: That is the generic Julia linear algebra functions which, by default on dense matrices, calls LAPACK, but on other matrices (like CuMatrix) it then dispatches to something else.

jarlebring: :+1:

jarlebring:
Okay. `lu!`+`ldiv!` seems to work in general. It doesn't work for `MMatrix` since `lu!` is not defined for that type but it could be easily be fixed with a dispatch on `ldiv_for_generated!`. For the moment, I will not add that since I'm unsure how important `MMatrix` is here.

ChrisRackauckas:
rename the other to `baseexp!`?

And do a test on the exponential integrators and this is good to go.

jarlebring:
> rename the other to `baseexp!`?

You mean `_exp_old!` -> `_baseexp!`? Does something speak against removing the old implementation? Less code to maintain. I don't see much advantage of the old code, except maybe that it is not generated.

jarlebring:
If we keep both, and  see the exponential integrators as the main application, I think we need the feature to specify which of the two non-allocating exp implementations should be used, eg here
https://github.com/SciML/ExponentialUtilities.jl/blob/5460b95594a9e23f418710e568db3a92d7bcfd26/src/krylov_phiv.jl#L85-L90

I think such features should only be added if they add value.

jarlebring:
When we are anyway messing with this code: `_exp!` is not exported but it is described on the main documentation page. How about renaming it to `exp_noalloc!` and exporting it?

ChrisRackauckas:
I think we mainly want to keep it around to potentially upstream it.

jarlebring:
Alright. Fine with me. Changed to `_baseexp!`. 

> And do a test on the exponential integrators and this is good to go.

But aren't they already tested already like in the test Phi?

README.md
```diff
@@ -90,10 +90,10 @@ For the other keyword arguments, `m` determines the dimension of the Krylov subs
 ## `_exp!`
 
 ```julia
-_exp(A)
+_exp!(A)
```
ChrisRackauckas, line 93: Oh I just noticed, the README should get an update too. Follow up PR?

jarlebring: Do you mean documentation of the kwargs in the README.md?

ChrisRackauckas: that no longer describes the method of `_exp!` and `_baseexp!` is not documented, right?

ChrisRackauckas: or is the new one based off of the 2008 paper?

jarlebring: They are both implementations of the Higham 2008 paper.

ChrisRackauckas: ahh okay, so all that's needed is a baseexp section then.

Issue #66, opened 2021-08-19
Handling of different exp methods


jarlebring:
There are several matrix exponential methods in the literature, suitable in different settings. At the moment we have Padé-based scaling and squaring. There is the direct diagonalization approach, but also polynomial based scaling and squaring (which might be better in a GPU setting), e.g. polynomial based of Sastre http://personales.upv.es/~jorsasma/software/expmpol.m or  https://arxiv.org/abs/2107.12198. If we add more implementations I think we need a way for a user to select this. Sketch of a system design using dispatch:

```julia 
struct ExpMethodHigham 
    do_balancing::Bool
end
ExpMethodHigham()=ExpMethodHigham(true);
struct ExpMethodDiagonalization
end
struct ExpMethodBase # Call Base.exp
end
```
....

Implementations could be done like this
```julia
function _exp!(A,method::ExpMethodDiagonalization,cache) 
    F=eigen!(A);
    copyto!(A,F.vectors*Diagonal(exp.(F.values))/F.vectors)
    return A
end
```

Functions that need exp can take the method object as a kwarg

```julia
function phi(z::T, k::Integer; cache=nothing,expmethod=ExpMethodHigham())
....
    P = _exp!(A,expmethod,cache)
```
The non-allocating can be handled with dispatch 
```julia
function allocate_mem(A,::ExpMethodHigham)
     return [similar(A) for i=1:5];
end
function allocate_mem(A,::ExpMethodDiagonalization)
     return nothing;
end
```
The function `allocate_mem` can either be called by the user or automatically if `caches=nothing`.

(We are working on other matrix exponential codes and this package would be a natural place to put them, otherwise we would need to create a separate package.)

ChrisRackauckas:
I think that makes a lot of sense.

PR #67, opened 2021-08-23
Selection of exponential in `exponential!`

jarlebring:
This is an implementation of #66.

The exponential function methods are  implemented in the function

`exponential!(A,method [cache])`

where method is an instantiation of one of the following types
* `ExpMethodHigham2005` the generated code, previously `_exp!`   (default)
* `ExpMethodHigham2005Base` previously `_exp_base!`
* `ExpMethodGeneric`  previously `exp_generic`
* `ExpMethodNative` calls `Base.exp`
* `ExpMethodDiagonalization` makes a diagonalization 

For example

``` 
julia> A=randn(50,50);
julia> method=ExponentialUtilities.ExpMethodNative();
julia> E1=exponential!(copy(A),method);
julia> E1 == exp(A)
true
```

Those functions that support preallocation handle  this by a call to `alloc_mem(A,method)`.

The name `exponential!` was the best function name I could come up with to replace `_exp!` and does not collide with `Base.exp`. Other suggestions welcome. 

This PR is not quite ready for merge. Hoping for some comments.


ChrisRackauckas (approved):

ChrisRackauckas:
Oh oops, didn't mean that. But it looks great. It's missing a deprecation warning for `_exp!` to `exponential!` though, since I know a few people were using that. Would you mind a follow up PR doing the deprecation? That would be easier than reverting.

jarlebring:
> Would you mind a follow up PR doing the deprecation? That would be easier than reverting

Sure.

andreasnoack:
@jarlebring What was the motivation for renaming `exp_generic` to `exponential!`? It still doesn't seem to mutate the argument so the exclamation mark is a bit misleading https://github.com/SciML/ExponentialUtilities.jl/blob/0a05727c747f98459f2789b988e320b9d422bf1a/src/exp_generic.jl#L116-L131

jarlebring:
`exponential!` is handling several implementations of the matrix exponential via dispatch on the algorithms specified in `method`. What was called `exp_generic` is now `exponential!(x,::ExpMethodGeneric)`. Although the `ExpMethodGeneric` doesn't modify the argument other algorithms do. To be more explicit we could have some algorithm implementations in `exponential(x,method)`  and some in `exponential!(x,method)`, but then I would like to have fallbacks such that `exponential!` always works. I didn't see much point at the time, but if you feel strongly maybe it can be considered. The fallback would be just something like
```
exponential!(args...) = exponential(args...)
```

andreasnoack:
A more standard approach would probably have been to introduce `exponential` as the general interface and then only define `exponential!` for the methods that actually mutates the arguments. 

More specifically, I had to change

```julia
ExponentialUtilties.exp_generic(A)
```
to
```julia
ExponentialUtilities.exponential!(
        A,
        ExponentialUtilities.ExpMethodGeneric(),
        nothing)
```
to upgrade to the new version of this package which seem like a step down.

Ideally, I could just use something like `ExponentialUtilities.exponential` and rely on the dispatch but that fails for `ForwardDiff.Dual` elements because of the very non-generic call to `LAPACK.gebal!`.

jarlebring:
> A more standard approach would probably have been to introduce exponential as the general interface and then only define exponential! for the methods that actually mutates the arguments.

Ok. Adding `exponential` now would be non-breaking so I think it can be done in a minor release. But removing `exponential!` for non-mutating implementations would break. 

> to upgrade to the new version of this package which seem like a step down.

Not sure what you refer to. Is `nothing` really needed? The fallback for the cache is to return `nothing`:

https://github.com/SciML/ExponentialUtilities.jl/blob/0a05727c747f98459f2789b988e320b9d422bf1a/src/exp.jl#L10-L13

> Ideally, I could just use something like `ExponentialUtilities.exponential` and rely on the dispatch but that fails for `ForwardDiff.Dual` elements because of the very non-generic call to `LAPACK.gebal!`.

The word "generic" is here used in the same sense as in `GenericLinearAlgebra`. You should use the Generic implementation if you have special number types, which it sounds like you have. So I think it is natural that you would need to specify a `method`.

andreasnoack:
> Is nothing really needed? The fallback for the cache is to return nothing

You are right. I just read the signature for the `exponential!` and assumed that `alloc_mem` would fail for my custom array.

> The word "generic" is here used in the same sense as in GenericLinearAlgebra. 

The original point with `GenericLinearAlgebra` is that you don't explicitly need dispatch to a special "generic" method. It simply overloads the usual methods and provide Julia fallbacks for whatever is needed. I.e. I'm not calling `svd(A, Generic())` when I compute the SVD of a `Matrix{Dual}`.

My point is just that we need a generic balancing method. There is one in GenericSchur.jl but it has a different API. Unfortunately, I haven't had time to implement one. IMO, the ideal should be that linear algebra code never need calls the LAPACK methods directly since it isn't (and shouldn't be) generic.

jarlebring:
>  I'm not calling svd(A, Generic()) when I compute the SVD of a Matrix{Dual}.

I understand that. Here we wanted to have many implementations of the exponential available. I don't see a better design decision than what we currently have.

> My point is just that we need a generic balancing method. [] haven't had time to implement one

Some reference implementations specify balancing but its value has been debated
Screenshot_20211123_140127
from https://blogs.mathworks.com/cleve/2012/07/23/a-balancing-act-for-the-matrix-exponential/#49528121-70e4-41ab-b759-e7463d083dd5 . Another solution to your problem is to remove it, or make it optional.

PR #68, opened 2021-08-24
Docs and deprecation for `exponential!`

### LINKED ISSUE #66: Handling of different exp methods

jarlebring:
Closes #66

### DIFF
--- a/src/exp_baseexp.jl
+++ b/src/exp_baseexp.jl
@@ -1,3 +1,10 @@
+"""
+    ExpMethodHigham2005Base()
+
+The same as `ExpMethodHigham2005` but follows `Base.exp` closer.
+
+"""
+
 struct ExpMethodHigham2005Base
 end
 function alloc_mem(A::StridedMatrix{T},method::ExpMethodHigham2005Base) where T<: LinearAlgebra.BlasFloat
--- a/src/exp_baseexp.jl
+++ b/src/exp_baseexp.jl
@@ -4,7 +4,6 @@
 The same as `ExpMethodHigham2005` but follows `Base.exp` closer.
 
 """
-
 struct ExpMethodHigham2005Base
 end
 function alloc_mem(A::StridedMatrix{T},method::ExpMethodHigham2005Base) where T<: LinearAlgebra.BlasFloat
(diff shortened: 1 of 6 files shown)


src/exp_baseexp.jl
```diff
@@ -1,3 +1,10 @@
+"""
+    ExpMethodHigham2005Base()
+
+The same as `ExpMethodHigham2005` but follows `Base.exp` closer.
+
+"""
+
 struct ExpMethodHigham2005Base
```
ChrisRackauckas, line 8:
```suggestion
"""
struct ExpMethodHigham2005Base
```

No space is required to attach the docstring to the struct.

ChrisRackauckas (approved):

PR #71, opened 2021-11-23
Call opnorm1 directly

andreasnoack:
Because `opnorm` branches on the value of `p`, inference runs on all the norms. This can be wasteful. Particularly, if a generic SVD method is loaded. This is avoided when `opnorm1` is called directly.


andreasnoack:
I think I've changed my mind here. This method is supposed to be generic so it shouldn't be using an unexported opnorm method that will be less likely to be defined for a custom type. Hence, we probably have to live with the inference overhead for now.

PR #72, opened 2021-11-23
Avoid integer overflow when scaling the argument in the generic exp

andreasnoack:
See the added test for an example.


ChrisRackauckas:
Thanks

Issue #81, opened 2022-04-03
Error using expv with Turing.jl


aelmokadem:
Hi,

I am running a Bayesian inference on an ODE model using Turing.jl. To try and speed up things, I represented the model as a linear system and expressed the coefficients in the form of a matrix to solve the system using matrix exponential (ExponentialUtilities.expv()). When I run the inference, I keep getting the error below (this is only the head of the error but I can try to share the rest if necessary). The error message is associated to the following line but I am guessing all other calls of `ExponetialUtilities.expv` would produce it as well:
```
u0_eoi = inv(K)*ExponentialUtilities.expv(durs[i], K, tmp_rate) - inv(K)*tmp_rate
```
Using `ExponentialAction.expv` works but very slow so I wanted to compare the performance to `ExponentialUtilities`! Ideally, I would be able to even use `ExponentialUtilities.expv_timestep` to get the full matrix without looping through timesteps!

Any ideas on how to solve this! Also, I would greatly appreciate it if you could share any alternative more efficient method to carry out the analysis with a linear solver!

Thank you.
Ahmed.

PS: the full model structure is also below. It is a PBPK model where I am modeling a drug's pharmacokinetics in patients. The drug is introduced via IV infusion so the model is set to calculate the state variables using the relation u(t)=A^-1 * exp(A*t) * rate - A^-1 * rate. Once infusion duration is over I switch to u(t)=exp(A*t) * u0 The CLint parameter includes random effects relevant to interindividual variability. The function PBPK() would take in the parameter set and generate the full model coefficients matrix. Within this function there is a line:
```
K = zeros(Base.promote_eltype(p[1]), ncmt, ncmt);
```
This line promotes the input parameter type when creating the `K` matrix. 
 I haven't attached the full function description but I can if necessary!

The error message:
```
[...]

PR #82, opened 2022-04-09
add Balance fallback via GenericSchur for non-BlasFloat

ChrisRackauckas:
Fixes https://github.com/SciML/ExponentialUtilities.jl/issues/81


RalphAS:
The permutations for the `balance!` version should be
```
   if ilo > 1       # apply lower permutations in reverse order
        for j in (ilo-1):-1:1; rcswap!(j, B.prow[j], X) end
    end
    if ihi < n       # apply upper permutations in forward order
        for j in (ihi+1):n;    rcswap!(j, B.pcol[j-ihi], X) end
    end
 ```

PR #84, opened 2022-04-21
Support CUDA arrays

GiggleLiu:
This PR will make time evolution in CuYao (the CUDA version of Yao, https://github.com/QuantumBFS/CuYao.jl) work properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.mem
 deps/deps.jl
 Manifest.toml
+.vscode
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -101,9 +101,10 @@ function arnoldi(A, b; m=min(30, size(A, 1)), ishermitian=LinearAlgebra.ishermit
     n = length(b)
     U = ishermitian ? real(T) : T
 
-    V = similar(A, T, (n, m + 1))
-    H = similar(A, U, (m+1, m))
-    fill!(H, zero(U))
+    # V stores the Krylov vectors
+    V = similar(b, T, (n, m + 1))
+    # H is a small dense array in tridiagonal form
+    H = zeros(U, (m+1, m))
 
     Ks = KrylovSubspace{T, U, real(T), typeof(V), typeof(H)}(m, m, false, zero(real(T)), V, H)
 
--- a/src/krylov_phiv.jl
+++ b/src/krylov_phiv.jl
@@ -123,11 +123,9 @@ function expv!(w::AbstractVector{Complex{Tw}}, t::Complex{Tt}, Ks::KrylovSubspac
         expH = exponential!(t * cache,expmethod)
         expHe = @view(expH[:, 1])
     end
-    lmul!(beta, mul!(w, @view(V[:, 1:m]), match_array_type(V, expHe))) # exp(A) ≈ norm(b) * V * exp(H)e
+    # `ArrayInterface.restructure` will convert the `expHe` to the target matrix type that can interact with `V`.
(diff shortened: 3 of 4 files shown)


GiggleLiu:
@ChrisRackauckas In case you did not see this PR.

.gitignore
```diff
@@ -3,3 +3,4 @@
 *.jl.mem
 deps/deps.jl
 Manifest.toml
+.vscode
```
YingboMa, line 6: BTW, you can set up global gitignore.

src/arnoldi.jl
```diff
@@ -101,9 +101,10 @@ function arnoldi(A, b; m=min(30, size(A, 1)), ishermitian=LinearAlgebra.ishermit
     n = length(b)
     U = ishermitian ? real(T) : T
 
-    V = similar(A, T, (n, m + 1))
-    H = similar(A, U, (m+1, m))
-    fill!(H, zero(U))
+    # V stores the Krylov vectors
+    V = similar(b, T, (n, m + 1))
+    # H is a small dense array in tridiagonal form
+    H = zeros(U, (m+1, m))
```
YingboMa, line 107: `H` will alway be `Matrix` if you don't use `similar`. Is that intended?

src/krylov_phiv.jl
```diff
@@ -123,8 +123,10 @@ function expv!(w::AbstractVector{Complex{Tw}}, t::Complex{Tt}, Ks::KrylovSubspac
         expH = exponential!(t * cache,expmethod)
         expHe = @view(expH[:, 1])
     end
-    lmul!(beta, mul!(w, @view(V[:, 1:m]), expHe)) # exp(A) ≈ norm(b) * V * exp(H)e
+    lmul!(beta, mul!(w, @view(V[:, 1:m]), match_array_type(V, expHe))) # exp(A) ≈ norm(b) * V * exp(H)e
 end
+# NOTE: this function is for CuArrays, which fall back to identity, i.e. keep the source a dense CPU array as it is.
+match_array_type(target::AbstractArray, source::AbstractArray) = source
```
YingboMa, line 129: Should this go to https://github.com/JuliaArrays/ArrayInterface.jl ?

src/arnoldi.jl
GiggleLiu: Yes, it is intended. This matrix is usually a very small one that can be easily diagonalized on CPU.
Also, there are a lot of element-wise operations to it.

.gitignore
GiggleLiu: Thanks for the tip, should I revert this file?

YingboMa: Nah, it's fine.

src/krylov_phiv.jl
GiggleLiu: Good point, I found the following API that will do this trick
```julia
julia> ArrayInterface.restructure(CUDA.randn(2,2),x )
2×2 CuArray{Int64, 2, CUDA.Mem.DeviceBuffer}:
 1  2
 3  4
```

GiggleLiu: Oops, this interface is not the desired one.

GiggleLiu: I reverted the change, and rename the `match_array_type` to `compatible_multiplicative_operand`.
Bceause it is not for matching certain array types, e.g. for sparse arrays, the operand is ok to be dense.

I do think this function should be in `ArrayInterface`, but that PR might require some work to understand different array libraries, let start by posting an issue there.

ChrisRackauckas:
Should we just merge this?

GiggleLiu:
> Should we just merge this?

That would be great!

PR #99, opened 2022-09-14
GPU-able inplace_add!

### LINKED ISSUE #89: GPU-Friendly `inplace_add!`
`inplace_add!` as currently implemented uses scalar indexing to update the diagonal of an array. Since `CuArray`s disallow scalar indexing outside the REPL, this method breaks in practice. 

The current implementation, reproduced from [here](https://github.com/SciML/ExponentialUtilities.jl/blob/4ff5442de0096d9ed65965ec3c0e3d696fa9b6e8/src/exp_noalloc.jl#L74-L79)
```julia
@inline function inplace_add!(A,B::UniformScaling) # Called from generated code
    s = B.λ
    @inbounds for i in diagind(A)
        A[i] += s
    end
end
```

SBuercklin:
Resolves #89 by adding a GPU-friendly codepath to `inplace_add!`. This could also be implemented as a different method and dispatch on `A`, but I would expect this to compile away the `if`. 

Verified on 1.8.1 with `CUDA@3.12.0`:
```julia
using CUDA, ExponentialUtilities

A_d = CUDA.randn(Float32, 1024, 1024)
eA_d = exponential!(copy(A_d), ExpMethodHigham2005(false));

A = collect(A_d);
eA = exp(A);

eA_GPU = collect(eA_d);

@assert eA_GPU ≈ eA
```


ChrisRackauckas:
Can you add a test? I'll setup GPU CI, but this needs a test script, a downstream Project.toml:

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/downstream/Project.toml

and a GPU test group.

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/runtests.jl#L158

If you make the test script I can handle the rest.

SBuercklin:
Added a test which verifies there's no scalar indexing happening for the `CuMatrix` `inplace_add!`, and a test which does the verification of the matrix exponential. It's written so we can add other exponential methods in the future if any of them happen to be GPU-compatible.

SBuercklin:
I didn't add it to `runtests.jl` though; `CUDA.jl` needs to be added as a dependency, so the coverage is likely still broken

src/exp_noalloc.jl
```diff
@@ -69,8 +69,12 @@ end
 # Inplace add of a UniformScaling object (support julia 1.6.2)
 @inline function inplace_add!(A, B::UniformScaling) # Called from generated code
     s = B.λ
-    @inbounds for i in diagind(A)
-        A[i] += s
+    if A isa GPUArraysCore.AbstractGPUArray
+        A .= A + s * I
```
ChrisRackauckas, line 73:
```suggestion
        A .= A .+ s * I
```

ChrisRackauckas:
It looks like it needs the buildkite script on master to run it, so let's see how this goes.

src/exp_noalloc.jl
ChrisRackauckas: Interesting https://buildkite.com/julialang/exponentialutilities-dot-jl/builds/3#01834a03-4610-4d15-a0d4-f66ebacd6e1c/183-347

PR #101, opened 2022-10-18
GPU support for expv and expv_timestep

albertomercurio:
Hello,

by changing only the syntax of several lines I implemented the GPU support for the `expv` and `expv_timestep` functions. The following code works now

```julia
using LinearAlgebra
using SparseArrays
using ExponentialUtilities
using BenchmarkTools
using CUDA
using CUDA.CUSPARSE
CUDA.allowscalar(false)
##
n = 70000
A = sprand(ComplexF64, n, n, 10 / n)
A = triu(A, 1) + sprand(ComplexF64, n, n, 1 / n)
b = rand(ComplexF64, n)

t = 0.1
ts = Array(LinRange(0, 1, 300))

expv(t, A, b)
@btime expv_timestep(ts, A, b)

A = CuSparseMatrixCSR(A)
b = CuArray(b)

expv(t, A, b)

@btime expv_timestep(ts, A, b)
```

### DIFF
--- a/src/krylov_phiv.jl
+++ b/src/krylov_phiv.jl
@@ -148,11 +148,8 @@ function ExponentialUtilities.expv!(w::GPUArraysCore.AbstractGPUVector{Tw},
         F = eigen!(SymTridiagonal(cache))
         expHe = F.vectors * (exp.(lmul!(t, F.values)) .* @view(F.vectors[1, :]))
     else
-        # lmul!(t, cache)
-        # expH = cache
-        # _exp!(expH)
-        # expHe = @view(expH[:, 1])
-        expH = exponential!(t * cache, expmethod)
+        lmul!(t, cache)
+        expH = exponential!(cache, expmethod)
         expHe = @view(expH[:, 1])
     end
 
@@ -243,7 +240,7 @@ function phiv!(w::AbstractMatrix, t::Number, Ks::KrylovSubspace{T, U}, k::Intege
         throw(ArgumentError("Cache must be a PhivCache"))
     end
     e, Hcopy, C1, C2 = get_caches(cache, m, k)
-    lmul!(t, copyto!(Hcopy, H[1:m, :])) # For the moment copyto! doesn't work with views. lmul!(t, copyto!(Hcopy, @view(H[1:m, :])))
+    lmul!(t, copyto!(Hcopy, @view(H[1:m, :])))
     fill!(e, zero(T))
     allowed_setindex!(e, one(T), 1) # e is the [1,0,...,0] basis vector
     phiv_dense!(C2, Hcopy, e, k; cache = C1) # C2 = [ϕ0(H)e ϕ1(H)e ... ϕk(H)e]
--- a/src/krylov_phiv.jl
+++ b/src/krylov_phiv.jl
@@ -132,12 +132,7 @@ end
 function ExponentialUtilities.expv!(w::GPUArraysCore.AbstractGPUVector{Tw},
                                     t::Real, Ks::KrylovSubspace{T, U};
                                     cache = nothing,
-                                    dexpHe::GPUArraysCore.AbstractGPUVector = typeof(w)(undef,
-                                                                                        Ks.m)) where {
-                                                                                                      Tw,
-                                                                                                      T,
-                                                                                                      U
-                                                                                                      }
+                                    expmethod = ExpMethodHigham2005()) where {Tw, T, U}
     m, beta, V, H = Ks.m, Ks.beta, getV(Ks), getH(Ks)
     @assert length(w)==size(V, 1) "Dimension mismatch"
     if cache == nothing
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Can you add a test in https://github.com/SciML/ExponentialUtilities.jl/blob/master/test/gpu/gputests.jl ?

src/krylov_phiv.jl
```diff
@@ -153,30 +148,31 @@ function ExponentialUtilities.expv!(w::GPUArraysCore.AbstractGPUVector{Tw},
         F = eigen!(SymTridiagonal(cache))
         expHe = F.vectors * (exp.(lmul!(t, F.values)) .* @view(F.vectors[1, :]))
     else
-        lmul!(t, cache)
-        expH = cache
-        _exp!(expH)
+        # lmul!(t, cache)
```
ChrisRackauckas, line 151: keep this so it's non-allocating?

src/krylov_phiv.jl
```diff
@@ -153,30 +148,31 @@ function ExponentialUtilities.expv!(w::GPUArraysCore.AbstractGPUVector{Tw},
...
+        # expH = cache
+        # _exp!(expH)
+        # expHe = @view(expH[:, 1])
+        expH = exponential!(t * cache, expmethod)
         expHe = @view(expH[:, 1])
     end
 
-    copyto!(dexpHe, expHe)
-    lmul!(beta, mul!(w, @view(V[:, 1:m]), dexpHe)) # exp(A) ≈ norm(b) * V * exp(H)e
```
ChrisRackauckas, line 163: is this extra allocation necessary?

src/krylov_phiv.jl
```diff
@@ -247,11 +243,11 @@ function phiv!(w::AbstractMatrix{T}, t::Number, Ks::KrylovSubspace{T, U}, k::Int
         throw(ArgumentError("Cache must be a PhivCache"))
     end
     e, Hcopy, C1, C2 = get_caches(cache, m, k)
-    lmul!(t, copyto!(Hcopy, @view(H[1:m, :])))
+    lmul!(t, copyto!(Hcopy, H[1:m, :])) # For the moment copyto! doesn't work with views. lmul!(t, copyto!(Hcopy, @view(H[1:m, :])))
```
ChrisRackauckas, line 246: what about `copy!`?

src/krylov_phiv.jl
albertomercurio: The other `expv` functions calls `exponential!(t * cache, expmethod)`, so I mantained the same structure. Moreover, this function requires the `expmethod` argument, and `_exp!` does not have any `expmethod` I think.

src/krylov_phiv.jl
albertomercurio: It needs to be a GPUArray type to perform the `lmul!` multiplication. But expHe is a CPUArray, so I  needed to move it into the GPU.

src/krylov_phiv.jl
albertomercurio: Also `copy!` seems not to work.

src/krylov_phiv.jl
ChrisRackauckas: but why not `lmul!` the `t*cache` to remove the array allocation?

src/krylov_phiv.jl
ChrisRackauckas: Why is `expHe` on the CPU? From 156 it should be a view of an array on the GPU?

src/krylov_phiv.jl
```diff
@@ -247,11 +243,11 @@ function phiv!(w::AbstractMatrix{T}, t::Number, Ks::KrylovSubspace{T, U}, k::Int
...
     fill!(e, zero(T))
-    e[1] = one(T) # e is the [1,0,...,0] basis vector
+    allowed_setindex!(e, one(T), 1) # e is the [1,0,...,0] basis vector
     phiv_dense!(C2, Hcopy, e, k; cache = C1) # C2 = [ϕ0(H)e ϕ1(H)e ... ϕk(H)e]
-    lmul!(beta, mul!(w, @view(V[:, 1:m]), C2)) # f(A) ≈ norm(b) * V * f(H)e
+    lmul!(beta, mul!(w, @view(V[:, 1:m]), typeof(w)(C2))) # f(A) ≈ norm(b) * V * f(H)e
```
ChrisRackauckas, line 250: Here too, why isn't C2 on the GPU in the first place?

src/krylov_phiv.jl
albertomercurio: Yes, I followed the concept of the Hessenberg matrix for the arnoldi iteration, which is alway a CPU matrix. So, I think that, since the krylov dimension is usually low, it was better to mantain that matrix on the CPU.

https://github.com/SciML/ExponentialUtilities.jl/blob/861bea82381a27c1e47249e9b272b0a19c9a2618/src/arnoldi.jl#L219-L226

What do you think?

src/krylov_phiv.jl
albertomercurio: Yes I will do that.

ChrisRackauckas: Since this allocates, can you put it behind an if statement to only apply to gpuarrays so that the CPU performance is unaffected, add a comment that links to that?

ChrisRackauckas:
In order to stop this from stagnating, add an if statement so that CPU dispatch doesn't regress, and let's merge. It fixes GPU support so that's better than nothing. But then open an issue pointing to this PR saying that the performance of some aspects can be improved.

albertomercurio:
Ok I restored the old `copyto!` with the views. The problem was that, when trying to make compatible with GPU Arrays, I canged the PhivCache to be similar to the GPU Array. But then I restored it to be always a CPU array, since it includes only small matrices, like the Hessenberg matrix of the Krylov subspace.

Thus, now there is no more this problem, and the original `copyto!` is restored.

ChrisRackauckas:
Thanks!

PR #107, opened 2022-10-30
add precompile script

daviehh:
The "time-to-first-plot" wait may sometimes add up; following the advice in [here](https://sciml.ai/news/2022/09/21/compile_time/), this PR adds some precompile scripts.

test w/
```
 @time [precomp_fx(; method, Tx) for method in precomp_ms for Tx in Txs];
```

before: (1st call)
> 29.416742 seconds (164.62 M allocations: 8.186 GiB, 7.63% gc time, 99.96% compilation time)

after precompile code & create_sysimage (1st call)
> 2.359652 seconds (12.56 M allocations: 821.236 MiB, 4.46% gc time, 99.73% compilation time)

2nd call is
> 0.117739 seconds (276.04 k allocations: 15.172 MiB, 5.51% gc time, 97.84% compilation time)

with test code:
```julia
using LinearAlgebra
using ExponentialUtilities

function precomp_gen_mat(; Tx, n)
    mx = rand(Tx, n, n)
    m = rand(Tx, n, n)
    v = rand(Tx, n)
    (mx, m, v)
end

function precomp_fx(; method, Tx, n=5)
    (mx, m, v) = precomp_gen_mat(; Tx, n)

    # mat exp
    cache = ExponentialUtilities.alloc_mem(mx, method)
    exm = exponential!(mx, method, cache)

    # exp v
    [expv(ts, f(m), v)
     for ts in (1.0, 1.0im)
     for f in (copy, Hermitian)
    ]

    # Subspace exponential caches not yet available for non-Hermitian matrices
    (Tx <: Real) && expv(1.0, Hermitian(m), v; mode=:error_estimate)

    # phi v
    # Subspace exponential caches not yet available for non-Hermitian matrices
    if Tx <: Complex
        [phiv(ts, Hermitian(m), v, 1)
         for ts in (1.0, 1.0im)
        ]
    else
        phiv(1.0, m, v, 1)
    end
end

precomp_ms = [
    ExpMethodHigham2005(),
    ExpMethodHigham2005Base(),
    ExpMethodGeneric(),
    ExpMethodNative(),
    ExpMethodDiagonalization()
]

Txs = [Float64, ComplexF64]
```


src/ExponentialUtilities.jl
```diff
@@ -34,6 +35,8 @@ include("krylov_phiv_adaptive.jl")
 include("kiops.jl")
 include("StegrWork.jl")
 include("krylov_phiv_error_estimate.jl")
+# precompile script
+include("ExpUtil_precomp.jl")
```
ChrisRackauckas, line 39: just name the file precompile.jl. No need to copy the package name in the package.

ChrisRackauckas:
Thanks!

PR #129, opened 2023-06-21
Doc improv

ArnoStrouwen:
Did SciML style change, @ChrisRackauckas ?
This is a larger diff than I expected.


ChrisRackauckas:
The formatter change is the big one.

ChrisRackauckas (approved):

Issue #130, opened 2023-07-06
Error: exponential! and exp_generic perform scalar indexing on gpu


arthur-bizzi:
Hey.

This happened as I tried to use the package for GPU-based matrix exponentiation (Base.exp can't do it). In constrast, `expv` works well. Though my julia-fu is not quite there yet, I would be happy to try and get this working if given some directions.
```
using CUDA, Flux, ExponentialUtilities

CUDA.allowscalar(false)

M = rand(Float32,10,10) |> gpu
v = rand(10) |> gpu 


ExponentialUtilities.exponential!(M) #ERROR: Scalar indexing is disallowed.
ExponentialUtilities.exp_generic(M) #ERROR: Scalar indexing is disallowed.
ExponentialUtilities.expv(1f0,M,v) #Works
```
The error message:
``` 
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:35
  [2] assertscalar(op::String)
    @ GPUArraysCore C:\Users\55619\.julia\packages\GPUArraysCore\uOYfN\src\GPUArraysCore.jl:103
  [3] getindex
    @ C:\Users\55619\.julia\packages\GPUArrays\t0LfC\src\host\indexing.jl:9 [inlined]
  [4] iterate
    @ .\abstractarray.jl:1220 [inlined]
  [5] iterate
    @ .\abstractarray.jl:1218 [inlined]
  [6] chkfinite(A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer})
    @ LinearAlgebra.LAPACK C:\Users\55619\AppData\Local\Programs\Julia-1.9.0\share\julia\stdlib\v1.9\LinearAlgebra\src\lapack.jl:84
  [7] gebal_noalloc!(job::Char, A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, scale::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer})
    @ ExponentialUtilities C:\Users\55619\.julia\packages\ExponentialUtilities\x24Py\src\exp_noalloc.jl:56
  [8] exponential!(A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, method::ExpMethodHigham2005, _cache::Tuple{Vector{CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}}, CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}})
[...]

PR #134, opened 2023-09-08
Fix GPU exponential! defaults

ChrisRackauckas:
Fixes https://github.com/SciML/ExponentialUtilities.jl/issues/130

### DIFF
--- a/test/gpu/gputests.jl
+++ b/test/gpu/gputests.jl
@@ -25,6 +25,8 @@ end
     eA = exp(A)
     A_d = cu(A)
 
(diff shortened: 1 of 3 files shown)


test/gpu/gputests.jl
```diff
@@ -25,6 +25,8 @@ end
     eA = exp(A)
     A_d = cu(A)
 
+    ExponentialUtilities.exponential!(A_d) # Make sure simple command works
```
ChrisRackauckas, line 28:
```suggestion
    exponential!(A_d) # Make sure simple command works
```

PR #137, opened 2023-09-16
docs 1.0 upgrade

ArnoStrouwen:


ArnoStrouwen:
@mortenpi is it expected behavior that Documenter complains about LinearAlgebra docstrings, while that module is not included in the modules keyword?
https://github.com/SciML/ExponentialUtilities.jl/actions/runs/6209285369/job/16856334243#step:5:28
https://github.com/SciML/ExponentialUtilities.jl/blob/master/docs/make.jl#L10

mortenpi:
You're overloading that function and attaching a docstring to it here:

https://github.com/SciML/ExponentialUtilities.jl/blob/f317b0655b74941ffee8a6cad89ea7312fc446c9/src/StegrWork.jl#L54-L67

So yes, it's expected. It's searching for docstrings (not functions etc) in your modules.

ChrisRackauckas:
Why is it not just the public functions? https://github.com/JuliaLang/julia/pull/50105 defines the public semantics, why wouldn't Documenter follow them?

mortenpi:
I mean.. that PR got merged like 2 days ago, and doesn't even exist in a Julia release. That said, yes, I do think we want to update Documenter's logic here eventually (https://github.com/JuliaDocs/Documenter.jl/issues/600#issuecomment-1722310282).

ChrisRackauckas:
Well you were saying you were going to throw a better error in the future. I was just saying that if anything should happen in the future, this case shouldn't error unless it's a public function.

mortenpi:
Ah. Well, you can still run into the error if you're legitimately extending a function from somewhere else. The confusion here, my perspective, is that it _looks like_ Documenter is complaining about someone else's docstring. But it's actually complaining about one of yours, which you do want to include in the manual (probably; if it's public stuff that is). So making this error better would be useful regardless.

That said.. `LAPACK.stegr!` is a public function though, no...? Or if it isn't, then why is ExponentialUtilities mucking with it? Or.. if it's adding like a private method, then the `public` keyword won't really help here, because you're running into the private method edge case that the new keyword doesn't really handle.

PR #147, opened 2023-12-05
Finish timestepping after happy breakdown

### LINKED ISSUE #143: `phiv_timestep` performance problems for scalar equations
**Describe the bug 🐞**

I consider this a bug since the performance difference is so severe, but I could move this to a feature request if desired.

In general, functions like `phiv_timestep` and `expv_timestep` are very fast for moderately-sized matrices. However, I'm noticing orders of magnitude in degradation, both in performance and in allocations, when the matrices are size-1.

It is possible for applications to circumvent this by manually implementing the functionality for scalar equations, but it would clearly be more convenient if the performance of these methods for scalar problems could be improved.

**Expected behavior**

Size-1 equations should be as fast as, or even faster than, larger ones.

**Minimal Reproducible Example 👇**

```julia
using ExponentialUtilities
function time(N)
        A = randn(N, N)
        b = randn(N, 1)
        @time phiv_timestep(1, A, b);
end

time(1)   # Warm up
time(1)   # 3.109664 seconds (18.85 M allocations: 18.841 GiB, 7.80% gc time)
time(2)   # 0.003811 seconds (3.81 k allocations: 406.266 KiB)
time(100) # 0.002961 seconds (261 allocations: 543.359 KiB)

```

**Environment (please complete the following information):**

  - Output of `using Pkg; Pkg.status()`

```julia
[d4d017d3] ExponentialUtilities v1.25.0
```

  - Output of `versioninfo()`

```julia
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake)
  Threads: 1 on 12 virtual cores
```

ArneBouillon:
Fixes #143 

I looked into that issue a bit more. Expokit.jl uses the same initial time step as ExponentialUtilities.jl, but does not face the dimension-1 issue since it stops timestepping upon a happy breakdown. When using the matrix dimension as the Krylov subspace dimension,  happy breakdown will occur and there is no need for taking (potentially very small) time steps.

### DIFF
--- a/src/krylov_phiv_adaptive.jl
+++ b/src/krylov_phiv_adaptive.jl
@@ -73,6 +73,10 @@ Niesen & Wright is used, the relative tolerance of which can be set using the
 keyword parameter `tol`. The delta and gamma parameters of the adaptation
 scheme can also be adjusted.
 
+When encountering a happy breakdown in the Krylov subspace construction, the
+time step is set to the remainder of the time interval since time stepping is
+no longer necessary.
+
 Set `verbose=true` to print out the internal steps (for debugging). For the
 other keyword arguments, consult `arnoldi` and `phiv`, which are used
 internally.
--- a/src/krylov_phiv_adaptive.jl
+++ b/src/krylov_phiv_adaptive.jl
@@ -128,9 +128,9 @@ function phiv_timestep!(U::AbstractMatrix{T}, ts::Vector{tType}, A, B::AbstractM
         P = similar(B, T, n, p + 2)         # stores output from phiv!
         V_tmp = similar(B, T, n, m + 1)
         H_tmp = fill(zero(T), m + 1, m)
-        Ks = KrylovSubspace{T, T, real(T), typeof(V_tmp), Matrix{T}}(m, m, false,
-            zero(real(T)), V_tmp,
-            H_tmp) # stores output from arnoldi!. Here we support also GPUs
+        Ks = KrylovSubspace{T, T, real(T), typeof(V_tmp), Matrix{T}}(
+            m, m, false, zero(real(T)), false, V_tmp, H_tmp
+        ) # stores output from arnoldi!. Here we support also GPUs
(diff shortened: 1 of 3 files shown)


ArneBouillon:
@ChrisRackauckas As far as I'm concerned, this one is ready for review. I added some documentation on this shortcut.

I'm unsure as to why some integration tests are failing. It seems unrelated to my changes.

src/krylov_phiv_adaptive.jl
```diff
@@ -174,6 +179,9 @@ function phiv_timestep!(U::AbstractMatrix{T}, ts::Vector{tType}, A, B::AbstractM
         end
         # Part 2: compute ϕp(tau*A)wp using Krylov, possibly with adaptation
         arnoldi!(Ks, A, @view(W[:, end]); tol = tol, m = m, opnorm = opnorm, iop = iop)
+        if Ks.wasbreakdown
```
ChrisRackauckas, line 182: It seems you're assuming `Ks.wasbreakdown` was false to start `phiv_timestep!` but that's not necessarily true if it's the same cache now reused for a new problem. I think this needs to be set to `false` at the top each time.

ChrisRackauckas:
See the note on re-solve. Also, does `expv_timestep!` need to be handled similarly?

ArneBouillon:
Good point, I didn't realise the caches were reused! `expv_timestep!` calls `phiv_timestep!` under the hood, so that doesn't need extra care.

ChrisRackauckas (approved):

ChrisRackauckas:
Thanks!

PR #154, opened 2023-12-24
switch to separate Project file for testing

ArnoStrouwen:
Are we committing to this testing style for all of SciML @ChrisRackauckas ?


ChrisRackauckas:
Does it work with the downgrade CI? I think it would be good to commit to it since I think the Pkg.jl folks want us to.

ArnoStrouwen:
I don't think the downgrade CI is working properly with this yet.
I've asked for clarification here:
https://github.com/cjdoris/julia-downgrade-compat-action/issues/3

ArnoStrouwen:
No longer relevant, `extras` is still the preferred testing method.

PR #157, opened 2024-01-04
Ci

ArnoStrouwen:
Let us use this one as a template for all the improvements you want @ChrisRackauckas.
I'm not sure where to put the caching action for Downstream and Downgrade CI.

PR #163, opened 2024-02-06
Update dependabot.yml

ArnoStrouwen:
I think it should be configured like this to only work on major releases.
Won't know for sure until typos does a next minor release.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,4 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "crate-ci/typos"
-        update-types: ["version-update:semver-patch"]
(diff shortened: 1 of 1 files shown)


ArnoStrouwen:
Is this a regression?
https://github.com/SciML/ExponentialUtilities.jl/actions/runs/7793155986/job/21252433155#step:6:131

ChrisRackauckas:
I'm not sure. I would think this just wasn't test on Windows much before?

ArnoStrouwen:
It passed that test here, IIRC:
https://github.com/SciML/ExponentialUtilities.jl/pull/157

ChrisRackauckas:
Peculiar

PR #170, opened 2024-03-13
ci: use reusable formatting workflow

thazhemadam:
Use organisation wide formatting workflow.

### DIFF
--- a/.github/workflows/FormatCheck.yml
+++ b/.github/workflows/FormatCheck.yml
@@ -1,4 +1,4 @@
-name: format-check
+name: "Format Check"
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
How's this?

thazhemadam:
There's still seems to be a few issues that I'm working out. But the general set up itself works.

ChrisRackauckas:
`Format Check / Format Checks / Comment Formatting Suggestions on PR (pull_request) Skipped`

Is this one supposed to show up as not run?

thazhemadam:
> `Format Check / Format Checks / Comment Formatting Suggestions on PR (pull_request) Skipped`
>
> Is this one supposed to show up as not run?

Yes. The "Comment Formatting Suggestions" step is run only if there are actually formatting suggestions that need to be commented. If the format checks are passing, then that job isn't run.

PR #185, opened 2024-12-10
Don't check for breakdown if adaptive=false

DanielVandH:
This changes `phiv_timestep` to no longer check for a happy-breakdown if `adaptive=false`.

## Checklist

- [ ] Appropriate tests were added

I'm not sure what example to come up with to get a new direct test for this. The only example I have is from private code.

- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ExponentialUtilities"
 uuid = "d4d017d3-3776-5f7e-afef-a10c40355c18"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>", "José E. Cruz Serrallés <Jose.CruzSerralles@nyulangone.org>"]
-version = "1.27.0"
+version = "1.27.1"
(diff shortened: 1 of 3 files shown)


DanielVandH:
Actually this seems to lead to a regression in https://github.com/SciML/ExponentialUtilities.jl/issues/143 since, without checking for breakdown, that example takes an extremely long time since the default is `adaptive=false`. Is there a better solution for this? Maybe checking that `adaptive=true` and simultaneously `tau` wasn't specified?

PR #196, opened 2025-11-19
Adapt expmv for AbstractGPUVector with complex t

fliingelephant:
The previous version does not work for the following code which does expmv for ```CuVector``` with complex t. Cause the dispatch for ```AbstractGPUVector``` only works for real t.
```Julia
using CUDA
using ExponentialUtilities
using Test

T = ComplexF64
v0 = randn(T, 4)
cuv0 = CUDA.cu(v0)
A = randn(T, 4, 4)
cuA = CUDA.cu(A)

Ks = ExponentialUtilities.arnoldi(A, v0; tol=1e-7, ishermitian=false, opnorm=1.0)
cuKs = ExponentialUtilities.arnoldi(cuA, cuv0; tol=1e-7, ishermitian=false, opnorm=1.0)

dt = 0.01im
ExponentialUtilities.expv!(v0, dt, Ks)
ExponentialUtilities.expv!(cuv0, dt, cuKs)
@test v0 ≈ collect(cuv0)
```
- Adapt ```expv!``` in ```src/krylov_phiv.jl``` for ```AbstractGPUVector``` with complex t
- Update testcases in ```test/gpu/gputests``` with complex t

PR #198, opened 2025-11-27
Fix GPU expv! to support complex t via shared implementation

ChrisRackauckas-Claude:
## Summary
- Extracts common GPU `expv!` logic into `_expv_gpu_impl!` helper function
- Provides separate method signatures for `t::Real` and `t::Complex` to avoid ambiguity with `AbstractVector` methods
- Avoids code duplication that was in PR #196

This addresses the review comment on #196 asking to use a Union instead of copy-pasting code. A direct Union doesn't work due to method ambiguities with the existing `AbstractVector` methods, but using a shared internal implementation achieves the same goal of avoiding code duplication.

## Test plan
- CI should pass all existing tests
- GPU tests with complex `t` should work as expected

### DIFF
--- a/test/gpu/gputests.jl
+++ b/test/gpu/gputests.jl
@@ -2,6 +2,7 @@ using LinearAlgebra
 using SparseArrays
 using CUDA
 using CUDA.CUSPARSE
+using ExponentialUtilities
 using ExponentialUtilities: inplace_add!,
                             exponential!, ExpMethodHigham2005, expv,
                             expv_timestep


ChrisRackauckas-Claude:
## Fix for GPU expv! with Complex t

The GPU tests were failing because the GPU `expv!` method signature was changed to require positional arguments without defaults:

```julia
# Was changed to this (broken):
function expv!(w::AbstractGPUVector, t, Ks, cache, expmethod)
```

When tests called `expv!(cuv0, dt, cuKs)` with 3 arguments, this didn't match the 5-argument GPU method, so Julia dispatched to the CPU Complex method instead, which doesn't properly adapt arrays for GPU.

**Fix:**
1. Renamed the implementation to `_expv_gpu_impl!` as an internal helper
2. Added two wrapper methods with proper keyword argument defaults:
   - `expv!(w::AbstractGPUVector{Tw}, t::Real, Ks; cache=nothing, expmethod=ExpMethodHigham2005Base())`
   - `expv!(w::AbstractGPUVector{Complex{Tw}}, t::Complex{Tt}, Ks; cache=nothing, expmethod=ExpMethodHigham2005Base())`

This maintains the shared implementation approach (no code duplication) while ensuring proper method dispatch for GPU arrays with both Real and Complex time parameters.

Also fixed a subtle issue: changed `exp.(lmul!(t, F.values))` to `exp.(t * F.values)` in the hermitian branch to avoid in-place mutation issues when `t` is complex.

PR #220, opened 2026-05-31
Re-enable downgrade CI

ChrisRackauckas-Claude:
**Please ignore until reviewed by @ChrisRackauckas.** Draft.

Re-enables the downgrade workflow (was disabled). The julia-actions/julia-downgrade-compat tooling problems that affected SciML monorepos are fixed and released in `@v2`; this package builds and passes its full test suite against the downgraded (minimal compatible) dependency versions. The runtest step uses `allow_reresolve: true` so the test environment can resolve transitive/test-only dependencies that the downgrade step does not lock.

Verified locally on Julia 1.11: downgrade resolve + build + `Pkg.test` all pass at downgraded versions.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -12,7 +12,6 @@ on:
       - 'docs/**'
 jobs:
   test:
-    if: false  # Disabled pending fix - see #208 for tracking re-enablement
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Redundant: master already re-enabled downgrade via the centralized strict @v1 workflow (#222/#224/#225). This branch's inline allow_reresolve:true would regress strictness. Closing.

PR #239, opened 2026-07-05
exp: Padé denominator solve via LinearSolve default algorithm choice

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft; opened by an agent at Chris's direction.

**Stacked on #236** (branches from it; shares the LinearSolve v4 dependency — merge #236 first).

First step of moving the package's remaining linear solves onto LinearSolve: `ExpMethodHigham2005Base`'s Padé denominator solve `(V−U) X = (V+U)` (both the small-norm and squaring branches) now goes through **LinearSolve's default algorithm choice** with the v4 batched matrix RHS, replacing the raw `LAPACK.gesv!` calls. Benefits: per-size/type algorithm selection (e.g. RecursiveFactorization when loaded), retcode-based failure instead of `LAPACKException`, consistency with `phi!`.

Verified locally against LinearSolve 4.0.0: `exponential! ≈ Base.exp` for n = 5, 40, 120 at small and large norms (exercising both converted sites).

Scoped small per review preference. Follow-ups, in order: (1) `ExpMethodHigham2005` (`exp_noalloc.jl`) with a workspace-embedded LinearSolve cache once the upstream in-place dense-LU option lands (SciML/LinearSolve.jl#1074 discussion); (2) benchmark whether a cached `init`/`solve!` here is worth the extra `alloc_mem` plumbing. Deliberately not converting `exp_generic.jl` (exotic-eltype path; generic LU is already optimal there) or the SMatrix `\` (container-preserving).

CI note: needs LinearSolve 4.0.0 registered in General (in flight) to resolve.


ChrisRackauckas-Claude:
Benchmarked honestly: **as written, this is a performance regression in the hot regime** — do not merge yet.

`exponential!` best-of-N, old (`gesv!`) vs new (`solve(LinearProblem(...))` per call), Julia 1.12, LinearSolve 4.0.0:

| n | norm | old | new | new + RF loaded |
|---|---|---|---|---|
| 10 | small | 10 µs | 17 µs | 18 µs |
| 30 | small | 81 µs | 96 µs | 88 µs |
| 60 | small | 358 µs | 561 µs | 454 µs |
| 60 | large | 534 µs | 792 µs | 568 µs |
| 120 | small | 3.96 ms | 3.52 ms | 4.00 ms |
| 250 | small | 9.96 ms | 9.32 ms | 9.80 ms |

Slower up to ~1.7× for n ≤ 60 (the Krylov-`Hm` sizes this package hits hardest), roughly neutral at n ≥ 120. The cost is the per-call `init` (A/b copies + default-alg dispatch), which `gesv!`'s single in-place LAPACK call doesn't pay; RecursiveFactorization narrows but doesn't close the gap.

Next revision here: thread a cached `init`/`solve!` LinearSolve cache through `alloc_mem` so repeated calls skip init and copies, then re-benchmark. If cached solve still can't match `gesv!` at n ≤ 60, the honest conclusion is that this solve should stay on `gesv!` (or the conversion should be behavior-only for the retcode benefit with an explicit perf note).

PR #240, opened 2026-07-05
exp: Padé denominator solve via LinearSolve default algorithm (alias_A)

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Draft; opened by an agent at Chris's direction.

Switches `ExpMethodHigham2005Base`'s cached Padé denominator solve from LinearSolve's default algorithm to **`GESVFactorization`** (SciML/LinearSolve.jl#1078, v4.3.0): one combined `gesv!` per refactorization, restoring raw-LAPACK parity at small n (**−1.8% vs the original `gesv!` code at n=5**, ties within noise at n=10–250; bit-identical accuracy and allocation bytes across all 12 benchmark cells). Also restores the historical `SingularException` contract — the default algorithm's QR safety-fallback silently resolved singular denominators.

Compat: `LinearSolve = "4.2"` — **unmergeable until LinearSolve 4.3.0 tags** (#1078). Full EU basictests pass locally against the dev'd branch (22 testsets, 1 pre-existing broken).

### DIFF
--- a/src/exp_baseexp.jl
+++ b/src/exp_baseexp.jl
@@ -16,11 +16,12 @@ function alloc_mem(
     U = Matrix{T}(undef, n, n)
     V = Matrix{T}(undef, n, n)
     temp = Matrix{T}(undef, n, n)
-    # Cached LinearSolve workspace for the Padé denominator solves, using
-    # LinearSolve's default algorithm choice (size-dependent: e.g. generic LU
-    # for tiny matrices, LAPACK/RecursiveFactorization above). The buffers are
-    # workspace-owned, so aliasing lets the factorization refactorize in place
-    # on every `solve!` instead of copying the n×n matrix.
+    # Cached LinearSolve workspace for the Padé denominator solves, using the
+    # default algorithm choice: it size-selects the fastest available LU
+    # (RecursiveFactorization when loaded — markedly faster than LAPACK at
+    # small/medium n). The buffers are workspace-owned, so aliasing lets the
+    # factorization overwrite Abuf in place on every `solve!` (no n×n copy),
+    # keeping each call allocation-free.
     Abuf = Matrix{T}(undef, n, n)
     Bbuf = Matrix{T}(undef, n, n)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
LinearSolve 4.3.0 (with `GESVFactorization`, SciML/LinearSolve.jl#1078) is released. Rebased onto current master and bumped compat to `LinearSolve = "4.3"`. Verified locally against 4.3.0: exp! ≈ Base.exp to 2.3e-15, exp-methods testset 40/40, Phi testset 284/284. Ready for CI once the registry propagates 4.3.0.

ChrisRackauckas-Claude:
Switched from `GESVFactorization` to LinearSolve's **default algorithm choice** (with `alias_A`/`alias_b`) after re-benchmarking with matched aliasing — thanks @ChrisRackauckas for pushing on this.

With `alias_A = true` the default:
- is allocation-competitive (no O(n²) backup copy — my earlier "+2–4.5%" was measured without alias on the default, an unfair comparison);
- has a **2–2.7× faster solve** at small/medium n when RecursiveFactorization is loaded (RFLU beats LAPACK LU — the common case in exponential-integrator stacks), vs GESV/LU which are pinned to LAPACK;
- needs only **LinearSolve 4.1** (alias_A in-place refactorization), not the 4.3-only `GESVFactorization` — so compat dropped `"4.3"` → `"4.1"` and this no longer waits on the latest release;
- auto-benefits from future LinearSolve algorithm improvements.

Singular-denominator behavior is unchanged: the solve returns a `Failure` retcode, which the code converts back to the `SingularException` `exp!` has always thrown.

Verified against **registered LinearSolve 4.1.0**: `exp!` ≈ `Base.exp` to ~2e-15, exp-methods testset 40/40, Phi testset 284/284. (`GESVFactorization` remains a useful general-purpose LinearSolve alg from #1078 regardless — it's just not the optimal pick for this fresh-every-call pattern.)

PR #251, opened 2026-07-15
Safeguard Arnoldi-seeded Krylov timesteps

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed

- apply the existing adaptive safety factor when the initial Krylov timestep is seeded from the Arnoldi norm estimate
- make the boundary-case regression deterministic with `Random.Xoshiro(14)`
- update the `phiv_timestep!` docstring to describe the initial-step behavior

## Why

PR #250 changed the matrix-free initial-step estimate. For the CI input, its first error estimate was narrowly accepted (`omega ≈ 1.16` with `delta = 1.2`) and the half-step relative error became `1.0152660196817718e-7`, just above the requested `1e-7` tolerance. The same safety factor already used by the adaptive controller brings that error to `6.6386e-9` without changing the test tolerance.

Bisecting identified `0d3f815` (PR #250) as the first bad commit; its parent produced a relative error of `1.5428773429973865e-8` for the same case.

## Local checks

- exact Julia 1.12.6 Core CI invocation: 676 passed, 1 pre-existing broken, 677 total
- deterministic regression before/after: `1.00753e-7` / `6.39e-9`
- 200-seed sweep: 0 failures; worst relative error `1.4357e-8`
- `julia +1.12.6 -m Runic --check --diff .`
- `git diff --check`

Please ignore this draft until it has been reviewed by @ChrisRackauckas.

### DIFF
--- a/src/krylov_phiv_adaptive.jl
+++ b/src/krylov_phiv_adaptive.jl
@@ -86,9 +86,9 @@ is estimated matrix-free from the Arnoldi Hessenberg built on the first Krylov
 step, so `opnorm(A, Inf)` is never evaluated: this needs no operator method
 beyond `mul!` -- suitable for matrix types (e.g. sparse GPU arrays) that do not
 support `opnorm` -- and reflects `A`'s action on the Krylov subspace the
-exponential actually explores. In this mode the initial `tau` defaults to the
-whole interval. Pass `opnorm` as a precomputed scalar (a norm or bound) or a
-function `opnorm(A, Inf)` to override the estimate with an explicit
+exponential actually explores. The first Arnoldi step supplies the scale used
+to seed the initial `tau`. Pass `opnorm` as a precomputed scalar (a norm or
+bound) or a function `opnorm(A, Inf)` to override the estimate with an explicit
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
Audit scratchpad for commit `b584d91ac88178124d599d80ee098c0680195b85`:

- Reproduced the exact seeded Core failure locally: relative error `1.0152660196817718e-7`.
- Bisected the first bad commit to `0d3f815` (#250).
- Added the deterministic regression and applied the algorithmic timestep safety factor.
- Exact local Core result: 676 pass, 1 pre-existing broken.
- A 200-seed local sweep completed with zero failures.
- Runic passed locally.
- Current PR result: all 35 checks are green, including Core/QA and the self-hosted GPU lane.

No tests were skipped, disabled, silenced, or loosened.

ChrisRackauckas-Claude:
Scheduled CI follow-up at 2026-07-15 05:50 EDT: all 35 checks are complete and successful, including the self-hosted GPU lane. No follow-up change was needed.

ChrisRackauckas (approved):

PR #255, opened 2026-07-17
Reuse exponential! workspaces across phiv!/expv! calls

ChrisRackauckas-Claude:
## Summary

**Stacked on #254** (includes its commit; review only the last commit here until #254 merges).

`phiv_dense!` (called by every Krylov `phiv!`) and the non-Hermitian branch of `expv!` called `exponential!` without a preallocated workspace, so every call re-allocated the `ExpMethodHigham2005Base` scratch matrices plus a LinearSolve Padé workspace for the (m+p)×(m+p) Hessenberg exponential. After #254, this was the dominant remaining per-step allocation in the Krylov-based exponential integrators.

`PhivCache` and `ExpvCache` now lazily hold `alloc_mem` workspaces — one per distinct extended-matrix size and exponential method requested, because a single integrator step evaluates phi functions of several orders (ETDRK4 alternates k=1/k=3; `phiv_timestep!` inside the EPIRK methods uses different column counts per stage), so a single-slot cache would be invalidated on almost every call. The store is bounded (reset past 16 entries) in case adaptive Krylov wanders over many subspace sizes. `phiv!` gains an `expmethod` keyword forwarded to `phiv_dense!` together with the workspace; matrix types without an `alloc_mem` preallocation (e.g. GPU) keep the previous two-argument `exponential!` path.

## Measurements (local, Julia 1.12.6)

Per call (n=400, m=30, k=3, preallocated caches): `phiv!` 90,128 → **1,152 bytes**; `expv!` → **752 bytes**.

Per step on the Kuramoto–Sivashinsky pseudospectral problem (N=128, non-allocating nonlinearity), combined with #254 (baseline = released v1.33.0):

| method | released | after |
|---|---|---|
| ETDRK2(krylov=true, m=20) | 114 KB | 3.0 KB |
| ETDRK4(krylov=true, m=20) | 405 KB | 12 KB |
| EPIRK4s3A | 636 KB | 11 KB |
| EPIRK5P2 | 981 KB | 16 KB |
| Exp4 | 1180 KB | 23 KB |
| Exprb32 | 250 KB | 3.7 KB |
| Exprb43 | 506 KB | 8.4 KB |

Caching-mode (non-Krylov) ETDRK methods were already 0 bytes/step and are unchanged.

## Tests

- New testset: alternating phi orders against one cache reuse exactly one workspace per size and match the uncached results; `expv!` workspace identity is stable across calls.
- Ran `Pkg.test()` locally: Core/basictests.jl 687 pass, 1 broken (pre-existing `@test_skip`).

---

**Note: please ignore this PR until reviewed by @ChrisRackauckas.**

https://claude.ai/code/session_01EDVGmmovzD5Aos3fFoPwMY

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,6 +22,7 @@ ExponentialUtilitiesStaticArraysExt = "StaticArrays"
 
 [compat]
 Adapt = "4.3"
+AllocCheck = "0.2"
 Aqua = "0.8"
 ArrayInterface = "7.1"
 DoubleFloats = "1.1.14"
@@ -41,6 +42,7 @@ Test = "1"
 julia = "1.10"
 
 [extras]
+AllocCheck = "9b6a8646-10ed-4001-bbdc-1d2f46dfbb1a"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 DoubleFloats = "497a8b3b-efae-58df-a0af-a86822472b78"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
@@ -51,4 +53,4 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Aqua", "DoubleFloats", "ForwardDiff", "Test", "SafeTestsets", "SciMLTesting", "StaticArrays", "Random"]
+test = ["AllocCheck", "Aqua", "DoubleFloats", "ForwardDiff", "Test", "SafeTestsets", "SciMLTesting", "StaticArrays", "Random"]
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,6 @@ ExponentialUtilitiesStaticArraysExt = "StaticArrays"
 
 [compat]
 Adapt = "4.3"
-AllocCheck = "0.2"
 Aqua = "0.8"
 ArrayInterface = "7.1"
 DoubleFloats = "1.1.14"
@@ -42,7 +41,6 @@ Test = "1"
 julia = "1.10"
 
 [extras]
-AllocCheck = "9b6a8646-10ed-4001-bbdc-1d2f46dfbb1a"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 DoubleFloats = "497a8b3b-efae-58df-a0af-a86822472b78"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
@@ -53,4 +51,4 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["AllocCheck", "Aqua", "DoubleFloats", "ForwardDiff", "Test", "SafeTestsets", "SciMLTesting", "StaticArrays", "Random"]
+test = ["Aqua", "DoubleFloats", "ForwardDiff", "Test", "SafeTestsets", "SciMLTesting", "StaticArrays", "Random"]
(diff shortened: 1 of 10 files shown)


ChrisRackauckas-Claude:
Both addressed in the latest commit.

**`Any` → concrete.** You were right — my "must be heterogeneous" reasoning was wrong. The workspace type is *size-independent* (`typeof(alloc_mem(30×30)) === typeof(alloc_mem(400×400))`, since LinearSolve's `DefaultLinearSolver` is a size-independent wrapper); it depends only on the element type and the method. So `expcache` is now `Vector{Tuple{Int, W}}` with `W` the concrete workspace type, recovered at construction via `Base.promote_op` (no allocation; `Nothing` for the fallback). `get_expcache!` now dispatches on the method type, so its return is concretely typed (`W` or `Nothing`, never a union).

**Why there were still allocations.** They were not the cache — they were entirely inside `ExpMethodHigham2005Base` itself (confirmed by comparison: the non-`Base` `ExpMethodHigham2005` was already 0 alloc/call). Two sources: `LAPACK.gebal!` allocated a balancing vector per call, and the Padé coefficient arrays `C = T[...]` were allocated fresh each call. Fixed by using `gebal_noalloc!` with a preallocated scale buffer, and `const` coefficient tuples fed through a type-stable function barrier.

Per-call (warm cache, Julia 1.12), released → now:

| | released | now |
|---|---|---|
| `phiv!` | ~90 KB | **0 B** |
| `expv!` | ~1 KB | 96 B |
| `phiv_timestep!` | ~3 KB | 352 B |

`phiv!` is now fully non-allocating. `exponential!(_, ExpMethodHigham2005Base, work)` itself goes 464 → 0 B/call, correct to machine precision across all norm ranges and Float64/ComplexF64/Float32. The remaining `expv!` 96 B (a column-view that does not elide) and `phiv_timestep!` 352 B (its own `[t]`/`coeffs` timestepping temporaries) are small, size-independent, and separate from both the cache and the exp method — I can chase them in a follow-up if you want them at zero too.

Local: Core 693 pass / QA 20 pass (1 pre-existing broken each).

ChrisRackauckas-Claude:
Rebased onto master. Two things worth flagging from the rebase:

**1. `GenericSchur.balance!` is a ~15 KB/call allocation regression on the CPU exp path.** Master (via the GPU-fallback work) replaced the LAPACK balancing in the `ExpMethodHigham2005`/`Base` matrix exponentials with `GenericSchur.balance!`. Measured on a 33×33 Hessenberg-sized matrix:

```
GenericSchur.balance! alone:        15168 bytes/call
exponential!(_, Higham2005Base):    15280 bytes/call
```

So every Krylov `phiv!`/`expv!`/`phiv_timestep!` step currently allocates ~15 KB from the reduced matrix exponential on master. This PR restores a non-allocating LAPACK `gebal_noalloc!` (writing the scale into the cache buffer) **scoped to the strided-BlasFloat CPU method only** — non-BlasFloat / GPU inputs keep the `GenericSchur.balance!` fallback, so GPU support is unaffected. With it, `exponential!(_, ExpMethodHigham2005Base, work)` is 0 bytes/call and the Krylov hot path is allocation-free. Happy to split this into its own PR if you'd prefer to review the balancing change separately.

**2. AllocCheck** (per request): wired in as a test dependency and a testset, but it **cannot** report a static zero here — `check_allocs` finds 346 sites for `phiv!` (even `arnoldi!` has 12), all tracing into `LinearSolve.init` (lazy first-call workspace build) and `LinearSolve.solve!`, which inlining attributes to the caller and shallow backtraces can't filter out. So the AllocCheck testset is a documented tracker (`broken = true`, the pattern OrdinaryDiffEqExponentialRK uses), and the **runtime size-independence testset is the authoritative allocation guard**.

Local: Core 695 pass / 3 broken (1 pre-existing + 2 AllocCheck markers); QA 21 pass.

PR #262, opened 2026-07-28
Update arnoldi.jl to include a reorthogonalisation option

cneverett:
Added an option to reorthogonalise (`reorth`) each basis vector during its construction in `arnoldi!`. This helps numerical roundoff from leading to poor basis orthogonalisation, even with a modified Gram-Schmidt procedure, especially when the Krylov subspace dimension is large. This can lead to better stability and smaller errors for large, highly stiff, non-normal systems.

This is my first time contributing to SciML, so I am not familiar with the format of pull requests. This pull request only adds a small feature and I have updated the function documentation to include it. I am not sure if it needs any specific tests.

## Checklist

- [ ] Appropriate tests were added
- [X] Any code changes were done in a way that does not break public API
- [X] All documentation related to code changes were updated
- [X] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [X] Any new documentation only uses public API

### DIFF
--- a/src/arnoldi.jl
+++ b/src/arnoldi.jl
@@ -285,8 +285,8 @@ Take the `j` 'th step of the Lanczos iteration.
 function arnoldi_step!(
         j::Integer, iop::Integer, A::AT,
         V::AbstractMatrix{T}, H::AbstractMatrix{U},
-        n::Int = -1, p::Int = -1,
-        reorth::Bool,
+        n::Int = -1, p::Int = -1;
+        reorth::Bool=false,
     ) where {AT, T, U}
     x, y = @view(V[:, j]), @view(V[:, j + 1])
     applyA!(y, A, x, V, j, n, p)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Do you have a test case where this was shown to be a tangible improvement?

cneverett:
Unfortunately I don't have a simple test case that shows the improvement. I have been developing a modified version of `arnoldi!` to work with very large systems that I have been modelling (stiff, non-normal Jacobians of size around 10^4x10^4 needing Krylov dimensions of around 256 to captures dynamics accurately) and I doing so I have found that adding in this simple re-orthogonalisation process does help reduce the `errest` of functions like `phiv!` used for time stepping and reduce noise in the output results. 

I can try to come up with a simple test-case to demonstrate these findings but it will take time. Or I could use the systems I am currently modelling and show some outputs with and without this re-orthogonalisation?

ChrisRackauckas:
Tests fail.

This should get at least one test exercising it.

cneverett:
I have now made `reorth` a kwarg in `arnoldi_step!` (set to `false` by default) so it shouldn't break downstream tests by changing the number of arguments. 

I will come up with at least one test demonstrating the difference, which will take a little time.

PR #276, opened 2026-08-26
ci: make Julia pre Core tests non-blocking (GenericLinearAlgebra clash)

ChrisRackauckas-Claude:
Split Julia `pre` Core matrix into a non-blocking group while GenericLinearAlgebra fails to precompile on Julia 1.13 due to LinearAlgebra stdlib method overwrite (upstream issue, not ExponentialUtilities).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/test_groups.toml
+++ b/test/test_groups.toml
@@ -12,8 +12,23 @@
 # where `using CUDA` errors with "CUDA driver not functional".
 
 [Core]
-versions = ["lts", "1", "pre"]
+versions = ["lts", "1"]
 os = ["ubuntu-latest", "macos-latest", "windows-latest", "macOS-15-intel"]
+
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Superseded by an upstream GenericLinearAlgebra fix rather than making Julia-pre non-blocking here:

https://github.com/JuliaLinearAlgebra/GenericLinearAlgebra.jl/pull/182

That PR stops overwriting LinearAlgebra's `eigencopy_oftype` / `eigvals!` for `UpperHessenberg` on Julia ≥ 1.13 (the method-overwrite precompile failure). Once it merges and registers as 0.4.1, Julia-pre Core here should be able to load again and this CI workaround can be closed.

Related partial fix: JuliaLinearAlgebra/GenericLinearAlgebra.jl#181 (eigencopy_oftype only).

PR #279, opened 2026-09-05
Diagonalize the Lanczos subspace matrix without allocating

ChrisRackauckas-Claude:
> **Ignore until reviewed by @ChrisRackauckas.**

`expT!` copies `α`/`β` into buffers held by `StegrCache` and diagonalizes them with a pure-Julia implicit-QL (Wilkinson-shift) eigensolver, `symtridiag_eigen!`, so every iteration of the error-estimate `expv!` allocates nothing. `StegrCache` gains three real workspace fields (`d`, `e`, `Z`); its constructor signature and the public API are unchanged.

This also fixes master's currently-red `QA (julia 1)` job. Since #278 that job fails ExplicitImports' `no_implicit_imports` on `eigen`; this PR removes the `eigen` call entirely. With #280 merged, QA is the only failing job left on master (https://github.com/SciML/ExponentialUtilities.jl/actions/runs/34133265541).

**Why not a preallocated LAPACK call:** the pre-#258 `StegrWork` path went through `@blasfunc`/`BlasInt`/`ccall`, and none of `LAPACK.stegr!`, `LAPACK.steqr!`, `LinearAlgebra.BlasInt`, `BLAS.@blasfunc`, or `USE_BLAS64` are `public` (`Base.ispublic` is `false` for all of them on 1.12), so strict QA removed it. A 30x30 tridiagonal QL is about 60 lines and works on cache buffers.

## Performance

Measured against the **preallocated** LAPACK path (v1.33.0's `StegrWork` `ccall` into `dstegr_` with a reused workspace), so both sides allocate zero bytes and the comparison is algorithm to algorithm. Julia 1.12.7, single-threaded BLAS, `@belapsed` medians, at the default subspace size 30:

| tridiagonal | QL (this PR) | preallocated LAPACK `stegr!` | allocating `eigen` (master) |
|---|---|---|---|
| random | 65 µs | 98 µs | 101 µs (16.7 kB) |
| Lanczos-generated | 53 µs | 109 µs | 112 µs (16.7 kB) |

The allocations `eigen` adds cost bytes but little time, about 3 µs at n = 100. The real gap is algorithmic, and it runs against the asymptotics: MRRR is `O(n^2)` where accumulating QL rotations is `O(n^3)`. LAPACK does win on generic tridiagonals past n of roughly 130, but not on the matrices this path actually produces, whose decaying off-diagonals and clustered Ritz values deflate in few QL sweeps and deepen MRRR's representation tree:

| n | random, QL/LAPACK | Lanczos, QL/LAPACK |
|---|---|---|
| 30 | 0.66x | 0.51x |
| 100 | 0.85x | 0.43x |
| 150 | 1.12x | 0.46x |
| 300 | 1.34x | 0.55x |
| 400 | 1.60x | 0.65x |

The docstring records that crossover so nobody raises `maxiter` into the losing regime unaware.

End to end against v1.33.0, the last release that still had the preallocated LAPACK path, `atol = rtol = 1e-12`, both at zero bytes on the dense case and agreeing to 12 digits:

| version | dense Hermitian n=300 | sparse Laplacian n=1e5 |
|---|---|---|
| v1.33.0 (preallocated LAPACK) | 1863 µs | 15305 µs |
| this PR (QL) | 1713 µs | 15159 µs |

So this is not only recovering what #258 lost; it is about 8% faster than the best the package has ever been on the dense case, and level where the matrix-vector product dominates.

## Allocation

Julia 1.12.7, seeded n = 300 Hermitian problem, measured inside a compiled caller:

```
=== master ===
  Ks.m=18  rel.err=4.0709797267623655e-13
  bytes per expv! call (n=300, m=30): 75072
  bytes per expT! call (j=30):        16072
=== this branch ===
  Ks.m=18  rel.err=4.069943031816314e-13
  bytes per expv! call (n=300, m=30): 0
  bytes per expT! call (j=30):        0
```

## Tests

New `Preallocated symmetric tridiagonal eigensolver` testset: eigenvalues match `eigvals`, `Z` is orthogonal and reconstructs the matrix, degenerate/diagonal/clustered cases, `expT!` matches dense `exp` for all four element types without mutating `α`/`β`, and `@allocated == 0` for both `symtridiag_eigen!` and `expT!`. The allocation assertions are the discriminating tests: on master `expT!` allocates 16072 bytes, and `symtridiag_eigen!` does not exist. Accuracy against `eigvals` is 1e-15 relative across sizes 5 to 200.

Local runs on f106365, rebased onto master a65c39a (Julia 1.12.7, all exit 0):

```
GROUP=Core   Core/basictests.jl |  852       1    853  4m20.9s
GROUP=QA     QA/qa.jl           |   28       2     30  3m05.9s   <- no errors; master has 1
docs/make.jl                    no errors
Runic --check / typos           clean
```

**Not verified:** GPU group (needs CUDA), the new 32-bit lane from #282, downstream, `julia pre` locally. GPU behaviour is unchanged by construction: `Ks.H` is always a host `Matrix`, so `expT!` never touches device memory. I confirmed with JLArrays that a device-resident `KrylovSubspace` gives identical results on master and this branch, and that the `expv(...; mode = :error_estimate)` convenience wrapper scalar-indexes on both (pre-existing, `_expv_ee` always builds a host subspace).

**Reviewer attention:** this replaces a vendored LAPACK routine with in-tree numerical code the package now maintains. The convergence bound is 30n QL sweeps per eigenvalue, mirroring LAPACK `steqr`; hitting it throws. Eigenvalues come back unsorted, which `expT!` does not need.

https://claude.ai/code/session_01LoGUEUrLDsXQjfRRTvYdya

### DIFF
--- a/src/krylov_phiv_error_estimate.jl
+++ b/src/krylov_phiv_error_estimate.jl
@@ -13,9 +13,9 @@ abstract type HermitianSubspaceCache{T} <: SubspaceCache{T} end
 Subspace-exponential cache for the error-estimate variant of [`expv!`](@ref)
 (the `mode = :error_estimate` path) on Hermitian operators with element type
 `T`. It is a concrete `HermitianSubspaceCache` sized for a Lanczos subspace of
-dimension up to `n`, and it uses the symmetric-tridiagonal eigensolver to
-compute the exponential of the tridiagonal subspace matrix on each iteration.
-Construct one directly, or let
+dimension up to `n`. Every iteration diagonalizes the tridiagonal subspace
+matrix with a preallocated implicit-QL eigensolver, so applying the subspace
+exponential allocates nothing. Construct one directly, or let
 [`get_subspace_cache`](@ref) build the right cache for a given
 `KrylovSubspace`.
 
@@ -33,6 +33,10 @@ A `StegrCache` sized for subspaces through dimension `n`.
   - `v::Vector{T}`: the subspace-propagated vector (length `n`) that holds the
     result of applying the subspace exponential.
   - `w::Vector{T}`: scratch vector (length `n`) for intermediate values.
+  - `d::Vector{R}`, `e::Vector{R}`: working copies (length `n`) of the
+    tridiagonal diagonal and off-diagonal, overwritten by the eigensolver, where
+    `R = real(T)`.
+  - `Z::Matrix{R}`: `n × n` eigenvector workspace.
 
 # Examples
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
CI result: 33 pass, 2 fail. Both failures are `Core (julia 1)` on ubuntu and windows in the pre-existing "exponential! on immutable (static) matrices" Jacobian testset (unseeded random `SMatrix`), which this PR does not touch and which also fails on master (https://github.com/SciML/ExponentialUtilities.jl/actions/runs/33955462765, https://github.com/SciML/ExponentialUtilities.jl/actions/runs/31863105429). Root cause is a Julia 1.12 miscompile of the StaticArrays product kernel for `Dual{_,Float32,16}` on AVX-512 runners; a separate PR with a workaround is coming.

Side effect worth knowing: master's `QA (julia 1)` job is also red since #278 because `eigen` is used in `expT!` without an explicit import (ExplicitImports `no_implicit_imports`). This PR removes that `eigen` call entirely, so QA passes here.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Claude Code 2.1.261, model: claude-fable-5-1) — https://claude.ai/code/session_01LoGUEUrLDsXQjfRRTvYdya

ChrisRackauckas-Claude:
Rebased onto master a65c39a (picks up #280 and the 32-bit lane from #282) and added a commit documenting the solver's size regime.

Corrected the performance claim in the description. The earlier "59 µs vs 100 µs" compared against the *allocating* `eigen` wrapper with the input copies inside the benchmark, which is not an algorithm-to-algorithm comparison. Re-measured against v1.33.0's preallocated `StegrWork` `ccall` path with zero allocations on both sides: QL is still faster at the sizes this path uses (53 µs vs 109 µs on a Lanczos tridiagonal at n=30), but it is `O(n^3)` against MRRR's `O(n^2)` and loses on generic tridiagonals past n of about 130. It keeps winning on Lanczos-generated ones at every size tested because their decaying off-diagonals deflate in few sweeps. Full tables in the description; the crossover is now in the `symtridiag_eigen!` docstring.

PR #282, opened 2026-09-07
Add 32-bit Core CI lane via test_groups.toml

ChrisRackauckas-Claude:
## Summary
- Add a `["Core 32-bit"]` matrix-only alias (`group = "Core"`, `arch = "x86"`, Linux only) so this package exercises 32-bit Julia via SciML/.github `@v1`.
- Infrastructure (arch axis + i386 libs) already lives in SciML/.github; this is the per-repo opt-in.

## Test plan
- [ ] CI shows a job like `Core (julia 1, ubuntu-latest, x86)`
- [ ] That job is green (or failure is a real Int32/WordSize bug to fix)


ChrisRackauckas-Claude:
Pushed Int64 parameter-search fix for the x86 InexactError in StaticArrays .

PR #286, opened 2026-09-18
Skip step-size adaptation after a Krylov happy breakdown

oameye:
When `arnoldi!` reports a happy breakdown the Krylov subspace is already `A`-invariant, so `_phiv!` is exact and increasing `m` leaves the error estimate bit-identical, which makes `kappa == 1` and turns `m + ceil(Int, log(omega / gamma) / log(kappa))` into `InexactError: Int64(Inf)`.


oameye:
Can you tag a release :)

